PR #1, opened 2015-02-10
Convert to Coffeescript

smizell:
Thoughts? Wrong move? Plain JavaScript wears me out now.

@fosrias @zdne


zdne:
I am fine with it! Was just wondering what is the motivation..

smizell:
1. Tired of writing JS
2. Curious
3. Want to extend some of these base types and prototypical inheritance is nasty

lib/primitives.js
```diff
@@ -1,218 +1,258 @@
-var _ = require('lodash');
+var ArrayType, BoolType, ElementType, KeyValueType, NullType, NumberType, ObjectType, StringType, convertFromDom, convertToType, _,
```
zdne: So I wonder – is this a standard procedure – bundling generated file with the package? Shouldn't this be generated during npm install?

smizell: I have not followed this, but should before merging. 

http://stackoverflow.com/a/13646337/4175777

zdne: See https://github.com/apiaryio/drafter.js/blob/master/package.json#L16 and https://github.com/apiaryio/drafter.js/blob/master/scripts/build

smizell:
@zdne Fixed

gulpfile.coffee
```diff
@@ -0,0 +1,23 @@
+jshint = require 'gulp-jshint'
+stylish = require 'jshint-stylish'
+mocha = require 'gulp-mocha'
+coffee = require 'gulp-coffee'
+gulp   = require 'gulp'
+
+gulp.task "lint", ->
```
zdne, line 7: Just out of curiosity: Why gulp instead of grunt?

zdne:
@smizell alzo note I can't merge this since it is your repo :)

gulpfile.coffee
smizell, line 7: Gulp works a ton better, as it is uses pipes to run tasks (which feels more like how Node stuff works), whereas Grunt used these crazy configuration objects.

PR #16, opened 2015-02-17
Add find method

smizell:


src/primitives.coffee
```diff
@@ -85,6 +85,19 @@ class Collection extends ElementType
...
 
+  findElements: (cond) ->
+    elements = @content.reduce (results, el) ->
+      return results.concat([el.findElements(cond)]) if el.elementType() in ['array', 'object']
+      return results.concat([el]) if cond(el)
+      return results
+    , []
+    _.flatten elements, true
+
+  find: (cond) ->
```
fosrias, line 94: Is this really how find should work? That is finds stuff with no context of where it was in the structure? Wondering the use of this? Further, seems recursion could solve this as well.

zdne, line 94: Yeah a comment would be good too :)

smizell, line 94: I changed how this works. Let me know if this is what you had in mind. The first was recursive, but just did not mutate any variables. This new way is recursive, too, but mutates, which is ok here I suppose.

@fosrias I think the function passed into this method should be able to receive the element, the parent element, and maybe depth? Currently, elements know of their children, but cannot know of their parent, which I think should change (additionally, an element could have multiple parents).

@zdne Would you like to see the style of comments used in Drafter? Has that worked well? I haven't wanted to comment much here as this is so early, it could change quite a bit. But I'd like to add something nice.

smizell:
This comment got buried: https://github.com/smizell/minim/pull/16#discussion_r24817738

src/primitives.coffee
zdne, line 94: @smizell 

> Would you like to see the style of comments used in Drafter? Has that worked well? I haven't wanted to comment much here as this is so early, it could change quite a bit. But I'd like to add something nice.

Any comments, regardless of the format are what I am looking for, even a human readable explanation will work :)

smizell, line 94: @zdne What kind of comment are you wanting? Something that explains the code, or explains the usage of the method? Or both?

fosrias, line 94: I am still wondering the value then of pulling out the raw values this way. What is the application of this IYO @smizell

smizell, line 94: I see. I think my tests were not descriptive, so I'll change that. This is creating an array of elements that pass the conditional function. It pulls out the entire element, not just the values. This is useful for searching for IDs or attributes in an entire document.

README.md
```diff
@@ -268,6 +268,17 @@ arrayType.push('d');
 console.log(arrayType.toValue()); // ['a', 'b', 'c', 'd']
 ```
 
+##### find
+
+The `find` method traverses the element tree and returns an `ArrayType` of all elements that match the conditional function given.
+
+```javascript
+var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
```
fosrias, line 276: BTW, I think we should not call these ArrayTypes but rather ArrayElements. In that way we are aligning our code with the refract domain. I can address this in the forthcoming PR where I implement the factory and break up the files.

smizell, line 276: Great point. I agree.

fosrias:
@smizell What is the status on this PR IYO?

PR #21, opened 2015-03-06
Errors

smizell:
This PR adds a special error type that is used to allow for chains to break, but not throw JavaScript errors. This way, you can write a client to handle what may be found in an object, yet not worry if that particular chain is not there.

Edit: this becomes useful when a chain fails on filters and maps.

``` coffeescript
foo = obj.get('bar').filter((el) -> el.has('baz')).map((el) -> el.toValue())
```

If `foo` above doesn't have `bar`, the error gets passed all the way through. This means there are no JavaScript errors thrown, but rather a minim `ErrorType` object that is returned on the first `get('bar')`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -330,3 +330,19 @@ var objectType = new minim.ObjectType()
   .set('email', 'john@example.com')
   .set('id', 4)
 ```
+
+### Errors
+
+Minim has its own error type that is passed down through the call chain. This allows clients to write chains that may break but do not throw JavaScript errors.
+
+```javascript
+var objectExample = {
+  foo: {
+    bar: 'value 1'
+  }
+};
+
+var objectType = minim.ObjectType(objectExample);
+var value1 = objectType.get('foo').get('bar'); // value of foo.bar
+var notFound = objectType.get('baz').get('foo') // Returns error object
+```
--- a/src/primitives.coffee
+++ b/src/primitives.coffee
@@ -27,8 +27,30 @@ class ElementType
 
   set: (@content) -> @
 
+  getProperty: -> new ErrorType 'Element does not have method getProperty', @
+
+  has: -> new ErrorType 'Element does not have method has', @
+
+  map: -> new ErrorType 'Element does not have method map', @
+
+  filter: -> new ErrorType 'Element does not have method filter', @
+
+  forEach: -> new ErrorType 'Element does not have method forEach', @
+
+  length: -> new ErrorType 'Element does not have method length', @
+
+  push: -> new ErrorType 'Element does not have method push', @
+
+  add: -> new ErrorType 'Element does not have method add', @
+
+  find: -> new ErrorType 'Element does not have method find', @
+
+  keys: -> new ErrorType 'Element does not have method keys', @
+
+  values: -> new ErrorType 'Element does not have method values', @
+
 class ErrorType
-  constructor: (@message = 'Unspecified error') ->
+  constructor: (@message = 'Unspecified error', @element) ->
 
   elementType: -> 'error'
 
@@ -53,6 +75,10 @@ class ErrorType
   add: -> @
 
   find: -> @
+
+  keys: -> @
+
+  values: -> @
 
 class NullType extends ElementType
   constructor: (attributes) ->
@@ -213,6 +239,7 @@ convertFromDom = (el) ->
   new NullType().fromDom el
 
 module.exports = {
+  ElementType
   ErrorType
   NullType
   StringType
--- a/src/primitives.coffee
+++ b/src/primitives.coffee
@@ -27,6 +27,33 @@ class ElementType
 
   set: (@content) -> @
 
+class ErrorType
+  constructor: (@message = 'Unspecified error') ->
+
+  elementType: -> 'error'
+
+  get: -> @
+
+  set: -> @
+
+  getProperty: -> @
+
+  has: -> false
+
+  map: -> @
+
+  filter: -> @
+
+  forEach: -> @
+
+  length: -> @
+
+  push: -> @
+
+  add: -> @
+
+  find: -> @
+
 class NullType extends ElementType
   constructor: (attributes) ->
     super 'null', null, attributes
@@ -186,6 +213,7 @@ convertFromDom = (el) ->
   new NullType().fromDom el
 
 module.exports = {
+  ErrorType
   NullType
   StringType
   NumberType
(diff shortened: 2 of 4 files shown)


smizell:
@fosrias This is kind of cool. We should chat about it soon. Code can be cleaned up quite a bit, but left it pretty raw for now.

smizell:
Here is what the Ion code looks like now:
https://gist.github.com/smizell/d90cf721f506f82e7d93

src/primitives.coffee
```diff
@@ -27,6 +27,61 @@ class ElementType
 
   set: (@content) -> @
 
+  getProperty: -> new ErrorType 'Element does not have method getProperty', @
+
+  has: -> false
+
+  map: -> new ErrorType 'Element does not have method map', @
```
fosrias, line 34: How about a `_noElement` function so you can do:

``` coffeescript
map: -> _noElement('map')
```

I think that would be a lot DRYer. Really need method missing functionality for this.

fosrias, line 34: or maybe `_methodMissing('map')`. That is probably semantically more correct.

test/primitives-test.coffee
```diff
@@ -146,6 +146,129 @@ describe 'Minim Primitives', ->
...
+      it 'returns an error', ->
+        expect(elType.map((el) -> 'test').elementType()).to.equal 'error'
+
+    describe '#filter', ->
+      it 'returns an error', ->
+        expect(elType.filter((el) -> true).elementType()).to.equal 'error'
+
+    describe '#forEach', ->
+      it 'returns an error', ->
+        expect(elType.forEach((el) -> console.log('test')).elementType()).to.equal 'error'
```
fosrias, line 173: console.log cruft?

README.md
```diff
@@ -330,3 +330,19 @@ var objectType = new minim.ObjectType()
...
+```javascript
+var objectExample = {
+  foo: {
+    bar: 'value 1'
+  }
+};
+
+var objectType = minim.ObjectType(objectExample);
+var value1 = objectType.get('foo').get('bar'); // value of foo.bar
+var notFound = objectType.get('baz').get('foo') // Returns error object
```
fosrias, line 347: So, I am still uncomfortable that I chain something and anywhere in that chain can be an error element and nothing tells me it is there. This feels very uncomfortable to me. It seems a greater evil to hide something wrong than to fail hard and you have to trap it or introduce a try method if you are concerned like:

``` coffeescript
object.try('get', 'baz').try('get', 'foo') 
```

or something that returns undefined if you want to gracefully do something. Maybe I a missing something, no pun intended.

test/primitives-test.coffee
```diff
@@ -146,6 +146,129 @@ describe 'Minim Primitives', ->
           foo: 'bar'
           z: 2
 
+  describe 'ElementType', ->
+    elType = undefined
+
+    before ->
+      elType = new minim.ElementType 'element'
+
+    describe '#getProperty', ->
```
fosrias, line 155: feels like this could really be DRYed out a lot:

``` coffeescript
describe 'ElementType', ->
  elType = new minim.ElementType 'element'
  missingMethods =
     getProperty: elType.getProperty('foo')
     map: elType.map((el) -> 'test')
     # Etc

  for method, missingMethod of missingMethods
    describe "##{method}", ->
      expect(missingMethod.elementType()).to.equal('error')
```

test/primitives-test.coffee
```diff
@@ -146,6 +146,129 @@ describe 'Minim Primitives', ->
...
+    describe '#keys', ->
+      it 'returns an error', ->
+        expect(elType.keys().elementType()).to.equal 'error'
+
+    describe '#values', ->
+      it 'returns an error', ->
+        expect(elType.values().elementType()).to.equal 'error'
+
+  describe 'ErrorType', ->
+    errType = undefined
```
fosrias, line 200: With some more creativity you can DRY all this out to an reuse the same missingMethods for both.

src/primitives.coffee
```diff
@@ -27,6 +27,61 @@ class ElementType
...
+
+  add: -> new ErrorType 'Element does not have method add', @
+
+  find: -> new ErrorType 'Element does not have method find', @
+
+  keys: -> new ErrorType 'Element does not have method keys', @
+
+  values: -> new ErrorType 'Element does not have method values', @
+
+class ErrorType
```
fosrias, line 52: Shouldn't toValue and other minim conversion methods be implement for polymorphism? Maybe they should raise an error or even something like raise @ and have ErrorType inherit from Error (that might be interesting). Then basically, you might solve my concern that when you try to use the ErrorType, it throws itself and gives a useful traced message or something.

src/primitives.coffee
smizell, line 34: JavaScript not having a method missing really stinks. There is no way to catch any method that is missing, so for this to actually work, there has to be some register that registers all potential methods before one is initialized. Not super cool, and wondering if it's not worth it to have it.

More on this later, though.

test/primitives-test.coffee
smizell, line 173: Not really. The console.log should never be called because `forEach` simply returns itself. I will change it because it may confuse.

README.md
smizell, line 347: So this really needs a couple of things (beside a method missing) to work well. Right now, the instance that failed actually gets passed into the error instance, so that an error instance knows what element failed and why it failed. 

But what would really help is if we kept track of where things were being called, so that an error could provide the path that failed.

So to sum it up, it does have some idea where the error happened, but it's not perfect.

test/primitives-test.coffee
smizell, line 155: Nice, will look at this.

src/primitives.coffee
smizell, line 52: Interesting idea. Wonder if that would work? Might be worth a shot.

smizell:
@fosrias See last commit. It throws/catches error when error type is initialized. Looks like it gives a full stack trace as well. Cool idea :)

test/primitives-test.coffee
smizell, line 200: Will DRY later, as I think I will be refactoring this in preparation for other element types.

test/primitives-test.coffee
smizell, line 155: Will DRY later, as I think I will be refactoring this in preparation for other element types.

smizell:
@fosrias Voting to close this for now per my thoughts here: #22.

smizell:
Not usable anymore.

PR #23, opened 2015-05-14
Update to latest spec

danielgtaylor:
This PR updates to the latest refract specification, as well as:
1. Convert to ES6 and remove Coffeescript dependencies
2. Use plain old npm as the build system
3. Adds and exposes a preconfigured type registry
4. Adds a `fromCompactRefract` method to read in compact serialized refract elements
5. Renames a few things, adds a few minor things (e.g. StringType#length)

It **does** handle plain js types for the various `push`, `add`, `set`, etc methods as well as already refracted types so long as they inherit from `ElementType`. It **does not** do any more work for error handling.

This is merging the work from apiaryio/fury.js#16 back into the minim project. The new SDK API in Fury.js is depending on these changes.

Sorry it's so large :weary:


README.md
```diff
@@ -142,7 +148,7 @@ This is a type for representing string values.
 The `get` method returns the value of the `StringType` instance.
 
 ```javascript
-var stringType = new minim.StringType('foobar');
+var stringType = new minim.StringType({}, {}, 'foobar');
```
danielgtaylor: The need to include meta/attributes in the constructor was mainly for consistency sake, but it makes for a worse experience in my opinion. Ideas?

smizell: One thing we could do is do this backwards:

``` javascript
var stringType = new minim.StringType('foobar', {}, {});
```

Where the first argument was the content of the element.

danielgtaylor:
Here's a case that this PR does not provide any conveniences for yet:

https://github.com/refractproject/refract-spec/blob/master/namespaces/mson-namespace.md#referencing--expansion

Edit: After spending some time with this today I think we may need to change how the type registry works. When loading an element type which hasn't been encountered before and is not present in the registry we should probably add it to the registry so that inheritance works. If `A` is an object and `B` is an `A`, then `B` should be loaded as an object and its content treated as a collection. With the code as-is in the PR it will ignore the content, which is a problem.

README.md
danielgtaylor: I've updated this in the latest commit. The order is now `ElementType(element, content, meta, attributes)` and the other types don't require an element, e.g. `StringType('some string');` just like before. The important note is that even the base element type has `content` as the second argument for consistency sake.

README.md
```diff
@@ -76,18 +82,18 @@ If the JSON above is used, it can be converted back to Minim types to make a rou
 var arrayType = minim.convertFromDom(aboveJson);
 ```
 
-### Element Methods
+### Element attributes
```
smizell: Capitalize "Attributes"

README.md
```diff
@@ -76,18 +82,18 @@ If the JSON above is used, it can be converted back to Minim types to make a rou
...
 
-Each Minim type provides the following the methods.
+Each Minim type provides the following attributes:
 
-#### elementType
+- element: `array` (string) - The name of the element type.
```
smizell: Should `array` be here?

README.md
```diff
@@ -206,6 +212,18 @@ var value = boolType.get() // get() returns 4
 
 This is a type for representing arrays.
 
+##### iteration
```
smizell: I'd capitalize this.

README.md
```diff
@@ -317,7 +335,37 @@ The `values` method returns an array of keys.
...
+```
+
+##### items
+
+The `items` method returns an array of key value pairs which can make iteration simpler.
+
+```js
+const objectType = new minim.ObjectType({ foo: 'bar' });
+
+for (let [key, value] of objectType.items()) {
```
smizell, line 361: :+1:

README.md
```diff
@@ -317,7 +335,37 @@ The `values` method returns an array of keys.
...
+
+### Element Registry
+
+Minim allows you to register custom types for elements. For example, if the element type name you wish to handle is called `category` and it should be handled like an array:
+
+```javascript
+var minim = require('minim');
+
+// Register your custom type
+minim.TypeRegistry.elementMap.category = minim.ArrayType;
```
smizell: Regarding this, it almost feels like there needs to be something for registering.

``` javascript
minim.TypeRegistry.register('category', minim.ArrayType);
```

index.js
```diff
@@ -1 +1,6 @@
-module.exports = require('./lib/primitives');
+var _ = require('lodash');
+
+module.exports = _.extend(
+  require('./lib/primitives'),
+  require('./lib/convert')
+);
```
smizell: Does this mutate what is exprted from `./lib/primitives`? Is that desired? It almost seems like you need:

``` javascript
module.exports = _.extend(
  {},
  require('./lib/primitives'),
  require('./lib/convert')
);
```

src/convert.es6
```diff
@@ -0,0 +1,53 @@
...
+}
+
+function getElementClass(element) {
+  let ElementClass = TypeRegistry.elementMap[element];
+
+  if (ElementClass === undefined) {
+    // Fall back to the base element. We may not know what
+    // to do with the `content`, but downstream software
+    // may know.
+    ElementClass = ElementType;
```
smizell: For the sake of trying not to mutate values, you could instead:

``` javascript
return ElementType
```

src/primitives.es6
```diff
@@ -0,0 +1,339 @@
...
+    this.content = content;
+
+    this[attributeElementKeys] = [];
+  }
+
+  toValue() {
+    return this.content;
+  }
+
+  toRefract(options={}) {
```
smizell: I think it would be good to start following this:
https://github.com/airbnb/javascript

If you want to try to do it in this PR, feel free, but we can also merge this and try to do it in another PR.

For example, this line should be:

``` javascript
toRefract(options = {}) {
```

smizell: Just to point it out, note the spacing around the equals sign.

src/primitives.es6
```diff
@@ -0,0 +1,339 @@
...
+    super('boolean', content, meta, attributes);
+  }
+}
+
+class Collection extends ElementType {
+  get length() {
+    return this.content.length;
+  }
+
+  get [Symbol.iterator]() {
```
smizell, line 156: Being relatively new to ES6, a comment here would help me know what's going on here.

src/primitives.es6
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  get(name) {
+    return name === undefined ? this : _.first(
+      this.content.filter((value) => value.meta.name === name)
+    );
+  }
+
+  set(name, value) {
+    const location = this.content.map(i => i.meta.name).indexOf(name);
```
smizell: Would something other than `i` be better here?

src/convert.es6
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/*
+ * Convert a long-form refract DOM into refract type instances. This uses
+ * the type registry above.
+ */
+export function convertFromRefract(dom) {
+  let ElementClass = getElementClass(dom.element);
+  return new ElementClass().fromRefract(dom);
+}
```
smizell: Just getting this idea out, but it may be good for us to come up with better names than "long-form" and "DOM" here. Something we stay consistent on throughout the spec and libraries.

smizell:
@danielgtaylor I have added a few comments throughout this PR, but I like what you'd done. This is all good stuff! We should try to get this merged ASAP once you have a look at the comments.

src/primitives.es6
```diff
@@ -0,0 +1,339 @@
...
+      }
+    });
+    return results;
+  }
+
+  find(condition, options) {
+    const newArray = new Collection();
+    newArray.content = this.findElements(condition, options);
+    return newArray;
+  }
```
smizell, line 268: I noticed in the reveal today, you used `findElements` instead of `find`. Are you wanting to move away from `find` or did `findElements` work better?

Just as a note, if you're wanting to move away from it, the README still shows `find` examples.

src/primitives.es6
```diff
@@ -0,0 +1,339 @@
...
+    this.content.push(convertToType(value));
+    return this;
+  }
+
+  add(value) {
+    this.push(value);
+  }
+
+  findElements(condition, options={}) {
+    const recursive = options.recursive === undefined ? false : options.recursive;
```
smizell: Do we trust the user, or do you want to convert the recursive value given to boolean?

``` javascript
!!options.recursive
```

Just a question that popped up.

src/primitives.es6
smizell, line 268: I'm wondering if `find` should be recursive by default, and then have something like `children` which is not recursive. What do you think?

src/primitives.es6
```diff
@@ -0,0 +1,339 @@
...
+
+  set(name, value) {
+    const location = this.content.map(i => i.meta.name).indexOf(name);
+
+    value = convertToType(value);
+    // TODO: Should we mutate or copy here? Of course it doesn't matter
+    //       for non-refracted elements as they get copied anyway, but
+    //       if the input is already refracted and we add it to multiple
+    //       objects with a different name suddenly we have a problem.
+    //       We have the same problem in the constructor above.
```
smizell, line 310: Good point. I'm wondering if we can handle this someway internally with a special kind of element wrapper. It would simply wrap an element in a property name.

Just throwing out ideas. Would be nice to have our cake and eat it, too.

src/primitives.es6
danielgtaylor: Ah, sorry, I'm coming from Python and PEP8 land :-)

smizell: I thought I saw some Python with the `for [key, value] in something.items()` :)

danielgtaylor:
@smizell I've addressed most of the feedback in the latest push. Please have a look and let me know if you have any other comments.

danielgtaylor:
Had to disable the `no-shadow` rule temporarily until eslint/eslint#2547

smizell:
:+1: Looks great!

PR #24, opened 2015-05-19
Upgrade eslint to fix no-shadow bug

danielgtaylor:
With the latest release of eslint and the inclusion of eslint/eslint#2545 we can now remove the technical debt from switching to a warning workaround for the `no-shadow` error.

https://github.com/eslint/eslint/releases

This is a minor change and shouldn't require a new release, but will help prevent style issues with future development.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -40,7 +40,7 @@
 /**
  * Variables
  */
-    "no-shadow": 1,                  // http://eslint.org/docs/rules/no-shadow
+    "no-shadow": 2,                  // http://eslint.org/docs/rules/no-shadow
     "no-shadow-restricted-names": 2, // http://eslint.org/docs/rules/no-shadow-restricted-names
     "no-unused-vars": [2, {          // http://eslint.org/docs/rules/no-unused-vars
       "vars": "local",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "minim",
-  "version": "0.3.0",
+  "version": "0.3.1",
   "description": "A library for interacting with JSON through Refract elements",
   "main": "lib/minim.js",
   "scripts": {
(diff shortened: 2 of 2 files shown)


smizell:
@danielgtaylor I think you will still need to bump to version 0.3.1 so we can push to NPM. Would you care to make that change on this PR?

---

Also, as a side note, I'd like to start trying something after reading [this](https://github.com/torvalds/subsurface/blob/master/README#L71-L114). I'm wanting to experiment with having a nicely-groomed commit history. Would you care to:
1. Only include one commit per PR. Feel free to include multiple commits in your PR for review, but rebase and squash them after review.
2. Use imperitave language, as mentioned in the link. This mainly means tell what the provided patch does.
3. Include a longer description of the commit in the commit message. This will mean the first line is a summary of the commit and the rest is a description. In the case of this PR, the first line of the commit message can be "Upgrade eslint to fix no-shadow bug" and the rest can be most of what you put in the description above minus the last line.

This is mostly an experiment. If you're up to it, I'll add this to the wiki for this project.

smizell:
So after talking with a few team members about it, instead of "one commit per PR," we should make it more general and liberal as "one commit per feature." A PR may have multiple commits, but a feature should only have one.

And let me know your thoughts :)

danielgtaylor:
Sure, the commit message thing is what I was doing on another project too and I generally liked it. My only concern is the "one commit per feature" thing. We should use merge commits for that in my opinion. Some features _will_ be better off with more than one commit and there's probably no avoiding it in the long run.

PR updated and should be ready to merge.

smizell:
@danielgtaylor Good feedback. What do you see as the balance? Should you try for one commit per feature and let multiple commits be for exceptions only?

danielgtaylor:
@smizell sure let's shoot for that. It should be pretty easy on smaller projects like this one.

smizell:
@danielgtaylor Great. I'm kind of testing this on the side to see if we could use this or see if others like it.

PR #27, opened 2015-05-25
Subclass Hooks

smizell:
This adds two methods on the base element class. One is for setting the element name and the other is for `attributeElementKeys`.

Additionally, I've added a `primitive` method for knowing what primitive type the sub class came from.

When reviewing this, it is probably helpful to go through each commit one by one. We can use parts of this functionality if it's too much of a departure from our current design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ The `map` method may be used to map over an array. Each item given is a Minim in
 ```javascript
 var arrayType = new minim.ArrayType(['a', 'b', 'c']);
 var newArray = arrayType.map(function(item) {
-  return item.elementType();
+  return item.element;
 }); // newArray is now ['string', 'string', 'string']
 ```
 
@@ -293,7 +293,7 @@ The `find` method traverses the entire descendent element tree and returns an `A
 ```javascript
 var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
 var numbers = arrayType.find(function(el) {
-  return el.elementType() === 'number'
+  return el.element === 'number'
 }).toValue(); // [1, 2, 3]
 ```
 
@@ -304,7 +304,7 @@ The `children` method traverses direct descendants and returns an `ArrayType` of
 ```javascript
 var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
 var numbers = arrayType.children(function(el) {
-  return el.elementType() === 'number';
+  return el.element === 'number';
 }).toValue(); // [3]
 ```
 
--- a/src/primitives.es6
+++ b/src/primitives.es6
@@ -10,6 +10,7 @@ import registry from './registry';
  * be converted from refract elements rather than simple types.
  */
 export const attributeElementKeys = Symbol('attributeElementKeys');
+export const storedElement = Symbol('storedElement');
 
 /*
  * ElementType is the base element from which all other elements are built.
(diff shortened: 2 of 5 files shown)


smizell:
cc @danielgtaylor

src/primitives.es6
```diff
@@ -33,7 +42,7 @@ export class ElementType {
   toRefract(options = {}) {
     const attributes = this.convertAttributesToRefract('toRefract');
     const initial = {
-      element: this.element,
+      element: this.elementName(),
```
JackuB: Idea: use getter/setter, so you can call `this.elementName` instead of `this.elementName()` since you have `this.meta`, `this.content` etc.

smizell: I would actually like to provide some method for `content`. The `meta` property is also something special, as it's a `Meta` class that has Minim instances for each key. I'm going to think about this more for now, but this is a good spot.

JackuB:
Found no issues :+1:

src/primitives.es6
```diff
@@ -17,13 +17,22 @@ export const attributeElementKeys = Symbol('attributeElementKeys');
...
     this.attributes = attributes;
     this.content = content;
+  }
+
+  elementName() {
+    return 'element';
+  }
 
-    this[attributeElementKeys] = [];
+  attributeElementKeys() {
```
danielgtaylor: FWIW, this was originally using a Symbol because it's considered private class data. For example, `Object.keys(myElement)` did not include the symbol. Now that it's a public method it will be listed there. Probably not a huge deal, but it's a side effect of this change to keep in mind.

src/primitives.es6
danielgtaylor: I'm :+1: for getters when we don't expect chaining, e.g. for the element name, meta, attributes, and content. Is there any reason **not** to use a getter?

README.md
```diff
@@ -95,6 +95,16 @@ Each Minim type provides the following attributes:
 
 Each Minim type provides the following the methods.
 
+#### elementName
+
+When called, this returns the name of the element. When subclassing a Refract primitive, create this method on the new class to set a new element name.
```
danielgtaylor: The [specification](https://github.com/refractproject/refract-spec/blob/master/refract-spec.md#element-object) calls this `element`. Maybe we should stick with that naming, or we should update the spec? I see minim as a reference implementation and keeping consistent naming is a good idea IMO.

src/primitives.es6
```diff
@@ -79,7 +88,6 @@ export class ElementType {
   }
 
   fromRefract(dom) {
-    this.element = dom.element;
```
danielgtaylor, line 82: Doesn't this break round-trips? For example:

``` js
minim.registry.register('foo', StringType);

minim.fromCompactRefract(['foo', {}, {}, 'test']).toCompactRefract();
// => ['string', {}, {}, 'test']
```

This means that I **must** create a subclass for each element now and can't just re-use existing ones with the same behavior.

smizell:
@danielgtaylor @JackuB You all have convinced me. I moved element to be a getter/setter and then moved the attribute keys as elements functionality back to constructor.

README.md
smizell: This should be addressed in [this](https://github.com/refractproject/minim/commit/31751160d1755e154b92b188afcd311a31889c7f) commit.

src/primitives.es6
smizell, line 82: I've added code and tests to accommodate for this.

smizell:
Would like to have these last three commits reviewed.
- [x] https://github.com/refractproject/minim/commit/31751160d1755e154b92b188afcd311a31889c7f
- [x] https://github.com/refractproject/minim/commit/f35b32abc40b24fe6c841deef0287d20fa1bffe3
- [x] https://github.com/refractproject/minim/commit/779393ee71dad3bd49ce5e1abe07f262aa347f9d

danielgtaylor:
:+1:

Issue #25, opened 2015-05-20
Convenience Methods


smizell:
In reading [this utility code](https://github.com/apiaryio/fury.js/blob/master/src/refract/util.es6), it may be helpful to provide convenience methods for better filtering. I think the utility code there is moving toward a structured way to query the tree, which I think we want to avoid at the moment. Using pure JS code is preferred. This means we need some better convenience methods.

For collections:
- [x] `.getById()` takes a string and returns the first element found with ID in entire tree
- [x] ~~`.filterByClass()` returns collection of elements where class matches given class name. Searches entire tree.~~

For items:
- [x] `.getClass()` returns an array of classes (or should it be collection?) (or `.classList()` or `.classes()`)
- [x] `.getId()` returns the ID of the element (undefined if not found) (or `.id()`)
- [x] `.withClass()` takes a string and returns true/false depending on if element has the class name given. Should also take an `options` argument where you can do things like `ignoreCase: true`.
- [x] `.hasClass()` returns true/false if element has classes
- [x] `.withId()` takes a string and returns true/false if the element has matching `meta.id`
- [x] `.hasId()` return true/false if element has an ID
- [x] `.withName()` takes a string and returns true/false if the element has matching `meta.name` (do we have a better name for this function?) Should take `options` like `hasClass`.
- [x] `.hasName()` returns true/false if name exists. Should take `options` like `hasClass`.
- [x] .`isType()` takes a string and returns true/false if the element is the given value (better name?)

To make this happen, I think it would be nice to also have some additional methods.
- [x] ~~`.getContent()` returns the content. Up for discussion... if it's array, it may be nice to return a collection element. This is people are not working directly with `content` property, which we should consider private.~~
- [x] ~~`.children()`with no condition returns a collection of of the content. This is essentially `.getContent()`~~
- [x] `.first()` returns `.get(0)`
- [x] `.second()` returns `.get(1)`

danielgtaylor:
I think overall this is a good idea. Some notes:
1. Do you want to add a `.last()` (i.e. `this.get(this.length - 1)`)?
2. I'm a little concerned about name collisions in the work I've been doing on [fury](https://github.com/apiaryio/fury.js/blob/master/src/refract/api.es6#L33-L47). There you can see that I'm adding convenience getters/setters for a bunch of stuff. Now I'm wondering if that's still the right approach as we add more and more to the element base classes.

smizell:
[...]

PR #28, opened 2015-05-25
Add Convenience Methods

smizell:
**Please do not merge**. This needs a rebase after #27 is addressed.

This adds several convenience methods while making meta into minim instances. This should address most of the items listed on #25.

cc @danielgtaylor

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,15 @@ var stringType = minim.convertToType("foobar");
 var compact = stringType.toCompactRefract(); // ['string', {}, {}, 'foobar']
 ```
 
+#### equals
+
+Allows for testing equality with the content of the element.
+
+```javascript
+var stringType = minim.convertToType("foobar");
+stringType.equals('abcd'); // returns false
+```
+
 ### Element Types
 
 Minim supports the following primitive types and the
@@ -309,6 +318,50 @@ var numbers = arrayType.children(function(el) {
 ```
 
 Because only children are tested with the condition function, the values `[1, 2]` are seen as an `array` type whose content is never tested. Thus, the only direct child which is a number type is `3`.
+
+##### getById
+
+Search the entire tree to find a matching ID.
+
+```javascript
+elTree.getById('some-id');
+```
+
+##### contains
+
+Test to see if a collection contains the value given. Does a deep equality check.
+
+```javascript
+var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
+arrayType.contains('a'); // returns true
+```
+
+##### first
+
+Returns the first element in the collection.
+
+```javascript
+var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
+arrayType.first(); // returns the element for "a"
+```
+
+##### second
+
+Returns the second element in the collection.
+
+```javascript
+var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
+arrayType.second(); // returns the element for "[1, 2]"
+```
+
+##### last
+
+Returns the last element in the collection.
+
+```javascript
+var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
+arrayType.last(); // returns the element for "3"
+```
 
 #### ObjectType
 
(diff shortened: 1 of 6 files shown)


smizell:
As a note, I added a special Meta class for handling meta attributes. I did it this way to lazy load each attribute because using a normal ObjectType lead to some crazy circular logic. It's also quite verbose at this point, but we can refactor later.

JackuB:
This PR should bump version in `package.json`.

Other than that seems good.

README.md
```diff
@@ -310,6 +319,50 @@ var numbers = arrayType.children(function(el) {
...
+
+Search the entire tree to find a matching ID.
+
+```javascript
+elTree.getById('some-id');
+```
+
+##### contains
+
+Test to see if a collection contains the value given.
```
danielgtaylor: Should elaborate on whether this is done via deep equality. E.g., does `arrayType.contains([1, 2])` return `true` or `false`?

README.md
```diff
@@ -310,6 +319,50 @@ var numbers = arrayType.children(function(el) {
...
+
+```javascript
+var arrayType = new minim.ArrayType(['a', [1, 2], 'b', 3]);
+arrayType.contains('a'); // returns true
+```
+
+##### first
+
+Returns the first item in the collection.
```
danielgtaylor: Does this return the element or the value? Item seems to imply element but it's not 100% clear to me from reading this.

src/primitives.es6
```diff
@@ -266,6 +342,34 @@ class Collection extends ElementType {
...
+   */
+  second() {
+    return this.get(1);
+  }
+
+  /*
+   * Return the last item in the collection
+   */
+  last() {
+    return this.get(this.get().length - 1);
```
danielgtaylor: Since a collection exposes a `length` getter you could use:

``` js
return this.get(this.length - 1)
```

src/primitives.es6
```diff
@@ -274,6 +378,16 @@ export class ArrayType extends Collection {
 
     super('array', converted, meta, attributes);
   }
+
+  contains(value) {
+    for (let item of this.content) {
+      if (item.toValue() === value) {
+        return true;
```
danielgtaylor, line 424: May want to consider some kind of deep equality, possibly as an option. Thoughts?

test/primitives-test.es6
```diff
@@ -2,6 +2,48 @@ import {expect} from './spec-helper';
...
+          class: ['a', 'b'],
+          name: 'name',
+          title: 'Title',
+          description: 'Description'
+        });
+      });
+
+      it('should initialize the correct meta data', () => {
+        expect(el.meta.id.get()).to.equal('foobar');
+        expect(el.meta.class.toValue()).to.deep.equal(['a', 'b']);
```
danielgtaylor, line 21: I wonder how we expect consumers of this API to know when to use `get` vs. `toValue`. Does it make sense to keep both?

src/primitives.es6
danielgtaylor, line 424: Also, what about objects? e.g. `new ObjectType({a: 1}).contains(1)` => `true`. Can also consider a `hasKey('a')` method.

test/primitives-test.es6
smizell, line 21: @danielgtaylor I have an issue open for this question. This is because we currently have two different ways `get()` works. Sometimes it returns `toValue()` and sometimes it returns `this`.

https://github.com/refractproject/minim/issues/30

src/primitives.es6
smizell, line 424: I've added code for this, both deep equality and `hasKey`. 

Now, `equals` also deep equality.

src/primitives.es6
smizell: Fixed.

README.md
smizell: Updated

README.md
smizell: Updated

test/primitives-test.es6
danielgtaylor, line 21: Ah I see, it's not that simple. Hmm.

danielgtaylor:
:+1:

PR #39, opened 2015-06-01
Add a name getter on meta.

danielgtaylor:
Looks like this was partially removed in a previous commit, but the setter
was still there. What do you think of adding it back in? If not, we can just
remove the setter as well.

This is still used by HTTP headers in the resource namespace, but I can add
the getter/setter in Fury instead if no other elements will use a name in
their meta.

cc @smizell

### DIFF
--- a/src/primitives.es6
+++ b/src/primitives.es6
@@ -59,6 +59,10 @@ export class Meta {
     this.setProperty('class', value);
   }
 
+  get name() {
+    return this.getProperty('name', '');
+  }
+
   set name(value) {
     this.setProperty('name', value);
   }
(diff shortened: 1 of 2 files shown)


smizell:
I almost think the header should now use the member element instead of name. Thoughts?

danielgtaylor:
Yeah we may want to do that for consistency. HTTP headers are kind of a special case object type though since both the keys and values _must_ be strings.

smizell:
Just realized too that name is no longer in meta in the refract spec.

smizell:
Closing this for now, as this is currently no longer in the Refract spec.

PR #42, opened 2015-06-05
ES5 Experiment

smizell:
I decided to spend a couple of hours today experimenting with moving the code to ES5. Here are some things I have found.
1. Using `_.extend` and `_.create` for inheritance is not that bad. The code doesn't look that far off from ES6 classes.
2. Modules are better in ES6.
3. I had to combine the registry and primitives into one file because of circular redundancy with `require`. 
4. Tests run in 0.5 seconds and it's so nice
5. Errors are for the right line numbers

Thoughts @danielgtaylor?


smizell:
Also, just a note, we need to talk about and decide on this route first. For the time being, this should not be merged.

smizell:
Did a test and ES6 classes still work as before:

``` js
import { StringType } from './lib/minim';

class MyString extends StringType {
  foo() {
    return 'bar';
  }
}

let test = new MyString('Hello World!');

console.log(test.primitive()); // string
console.log(test.toValue()); // Hello World!
console.log(test.foo()); // bar
```

.eslintrc
```diff
@@ -1,171 +1,26 @@
...
+        ],
+        "semi": [
+            2,
+            "always"
+        ]
+    },
+    "env": {
+        "node": true,
+        "browser": true
+    }
```
danielgtaylor, line 25: I think we should probably follow the ES5 version of the [Airbnb style guide](https://github.com/airbnb/javascript/tree/master/es5). I don't see an eslintrc for it anywhere - did you make this one based on those rules?

smizell, line 25: I just did the eslint init. We definitely need something better!

lib/base.js
```diff
@@ -0,0 +1,790 @@
...
+    this.content = content;
+    return this;
+  },
+
+  equals: function(value) {
+    return _.isEqual(this.toValue(), value);
+  }
+});
+
+Object.defineProperty(ElementType.prototype, 'element', {
```
danielgtaylor, line 308: Is there any way to do this within the `extend` syntax above? This feels a little out of place, but it could work okay.

lib/base.js
```diff
@@ -0,0 +1,790 @@
...
+});
+
+function Collection() {
+  ElementType.apply(this, arguments);
+}
+
+// TODO: Iterators for ES5
+// get [Symbol.iterator]() {
+//   return this.content[Symbol.iterator];
+// }
```
danielgtaylor, line 380: I guess we'll need to check if `Symbol` is even available, then attach this when it's possible. Probably need to play with this a bit.

package.json
```diff
@@ -1,23 +1,19 @@
...
+  "version": "0.6.0",
   "description": "A library for interacting with JSON through Refract elements",
   "main": "lib/minim.js",
   "scripts": {
-    "lint": "eslint --ext .js,.es6 src",
-    "precompile": "npm run lint",
-    "compile": "babel --optional runtime src --out-dir lib",
-    "precoverage": "npm run compile",
-    "coverage": "istanbul cover _mocha -- --compilers es6:babel/register -R spec --recursive",
+    "lint": "eslint --ext .js lib",
```
danielgtaylor: `--ext` defaults to `js` so this can probably be removed.

test/primitives-test.js
```diff
@@ -1012,104 +1026,104 @@ describe('Minim Primitives', () => {
...
+    // describe('#[Symbol.iterator]', function() {
+    //   it('can be used in a for ... of loop', function() {
+    //     var items = [];
+    //     for (let item of objectType) {
+    //       items.push(item);
+    //     }
+    //
+    //     expect(items).to.have.length(2);
+    //   });
+    // });
```
danielgtaylor, line 1110: I wonder if it makes sense to include an ES6 test (or a few) to ensure compatibility? We could either try to detect e.g. iojs 2+ or use babel (but only on those specific tests). Thoughts?

danielgtaylor:
@smizell overall I think this looks fine. Do you know if `super` still works as expected from ES6 (both in constructors and other methods)?

Issue #34, opened 2015-05-29
Sync code with spec


smizell:
It would be good to sync up the code of Minim with the spec.
- [x] Change classes from `StringType` to `String Element`
- [x] Remove instances of DOM
- [ ] Support Extend
- [ ] Support Select/Option

danielgtaylor:
We also need to support `extends`, `select`, `option` elements. [Spec link](https://github.com/refractproject/refract-spec/blob/master/namespaces/mson-namespace.md#one-of)

smizell:
@danielgtaylor I've updated the checklist for this.

PR #45, opened 2015-06-12
Sync with Refract Spec

smizell:
This handle issue #34.

This is kind of... nasty. But it aligns the code and terminology used through with the spec. This breaks every possible interface for the module :)

cc @danielgtaylor

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Refract is a JSON structure for JSON documents to make a more flexible document 
 1. Name of the element
 1. Metadata
 1. Attributes
-1. Content (which can be of different types depending on the element)
+1. Content (which can be of different elements depending on the element)
 
 An element ends up looking like this:
 
@@ -36,12 +36,12 @@ var el = {
 
 ## Usage
 
-### Converting to Types
+### Converting Javascript Values into Elements
 
 ```javascript
 var minim = require('minim');
-var arrayType = minim.convertToType([1, 2, 3]);
-var refract = arrayType.toRefract();
+var arrayElement = minim.convertToElement([1, 2, 3]);
+var refract = arrayElement.toRefract();
 ```
 
 The `refract` variable above has the following JSON value.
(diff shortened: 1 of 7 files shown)


zdne:
:boom: :boom: :boom: :boom:   :dancers:

README.md
```diff
@@ -36,12 +36,12 @@ var el = {
 
 ## Usage
 
-### Converting to Types
+### Converting to Elements
```
danielgtaylor: I feel like maybe this is more clear if we mention both the `to` and `from`:

Converting Javascript Values into Elements

README.md
```diff
@@ -74,12 +74,12 @@ The `refract` variable above has the following JSON value.
 }
 ```
 
-### Converting from Types
+### Converting from Elements
```
danielgtaylor: Converting Refract JSON into Elements
Converting Serialized Refract into Elements

README.md
smizell: I think you are 100% correct

README.md
```diff
@@ -74,12 +74,12 @@ The `refract` variable above has the following JSON value.
...
 
-If the JSON above is used, it can be converted back to Minim types to make a roundtrip.
+If the JSON above is used, it can be converted back to Minim elements to make a roundtrip.
 
 ```javascript
-var arrayType = minim.convertFromDom(aboveJson);
+var arrayElement = minim.convertFromElement(aboveJson);
```
danielgtaylor: In this method the word `element` seems to refer to a JSON serialization of refract, while above `convertToElement` uses the word `element` to mean an instance of an element class. I wonder if we can't make this a little more clear / less ambiguous.

``` js
var arrayElement1 = minim.fromValue([1, 2, 3]);
var refracted = arrayElement1.toRefract();
var arrayElement2 = minim.fromRefract(refracted);
```

Edit: the individual element classes use `toElement` to convert from values, so maybe that part makes sense. I don't know - trying to think of this from a newcomer checking out the library.

lib/base.js
```diff
@@ -87,22 +87,22 @@ var TypeRegistry = createClass({
...
   },
 
   /*
-   * Convert a long-form refract DOM into refract type instances.
+   * Convert a long-form refract document into refract element instances.
    */
-   fromRefract: function(dom) {
-    var ElementClass = this.getElementClass(dom.element);
-    return new ElementClass().fromRefract(dom);
+   fromRefract: function(doc) {
```
danielgtaylor, line 99: :+1: doc over dom

danielgtaylor:
Generally I agree with this change, just a few minor points above!

smizell:
@danielgtaylor I've addressed comments. I'd like to squash some of these before merging, so let me know if it's good and I'll merge. Thanks!

danielgtaylor:
Other than that comment above I think this looks good. We'll need to take some time to update the downstream libraries like Fury.

smizell:
> We'll need to take some time to update the downstream libraries like Fury

I'm going to be working on this a lot over the next few days. I'm wondering if it makes sense to wait until I'm finish (which will be the end of our sprint next Tuesday).

smizell:
Also, I sent out a document about some proposed changes for the API Description namespace. I think it would be good to wait on that, too, and fix it all at once.

Issue #40, opened 2015-06-01
Meta getProperty mutates underlying storage


danielgtaylor:
This leads to non-obvious behavior which I think we need to properly define or change. Here is the implementation as it stands:

[pimitives.es6#34-40](https://github.com/refractproject/minim/blob/master/src/primitives.es6#L34-L40)

``` js
class Meta {
  getProperty(name, value) {
    if (!this.meta[name]) {
      this.meta[name] = registry.toType(value);
    }

    return this.meta[name];
  }
}
```

Here is the surprising part:

``` js
const element = new ElementType();

console.log(element.toCompactRefract());
// => ['element', {}, {}, null]

// Access the non-existing name property
element.name;

// What? Now there is a meta name for some reason!?
console.log(element.toCompactRefract());
// => ['element', {name: ''}, {}, null]

// Try to get something from meta with a default if not found
element.getProperty('foo', 'bar');

// What? Now my default value is getting serialized!
console.log(element.toCompactRefract());
// => ['element', {name: '', foo: 'bar'}, {}, null]
```

I'm sorry I glossed over this in the initial review! I believe the code is written this way to cache the conversion to refract element classes for the default value, but I think we need to either do that conversion each time or keep a separate cache that doesn't mutate the original meta values..

Thoughts @smizell?

smizell:
@danielgtaylor Great find here. I think you're right. I'll be thinking on this as I do not think what I currently wrote is helpful to the user. I also think I should have made `getProperty` a private method. 

I'll be thinking on this. I'm working on a proposal for Minim that I'll be running by you soon.

PR #46, opened 2015-06-17
Add convenience methods for meta attributes

smizell:
This allows you to access meta attributes with something like:

``` js
element.id
```

Instead of:

``` js
element.meta.id
```

Note: There is still more work to fix this.
1. There is a bug with this, as it still mutates `meta` #40 
2. It is not DRY, as it has copy/paste code from Meta class

I will try to address these together as I refactor how we use `meta`.

### DIFF
--- a/lib/base.js
+++ b/lib/base.js
@@ -160,7 +160,7 @@ var Meta = createClass({
       return this.getProperty('class', []);
     },
     set: function(element) {
-      this.setProperty('id', element);
+      this.setProperty('class', element);
     }
   },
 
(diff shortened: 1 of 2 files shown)


smizell:
cc @danielgtaylor

lib/base.js
```diff
@@ -315,6 +327,53 @@ BaseElement = createClass({
...
+      this.setMetaProperty('id', element);
+    }
+  },
+
+  class: {
+    get: function() {
+      return this.getMetaProperty('class', []);
+    },
+    set: function(element) {
+      this.setMetaProperty('id', element);
```
fosrias: C&P error? Should be `class`, not `id` I think.

fosrias: Should have test that actually catches this.

smizell: Yeah, will address in future PR, which will be for next Minim release.

smizell: Fixed

smizell: (nice catch, btw)

fosrias:
@smizell Looks good. Merge when ready.

fosrias:
:+1:

PR #48, opened 2015-06-18
Update Interface

smizell:
This PR does several things.
1. Allows `obj.set({ foo: 'bar' })` to work for objects
2. Converts meta to being a Minim object
3. Converts attributes to being a Minim object
4. Adds new `getValue` method which is short for `obj.get('foo').toValue()`. Now you can do `obj.getValue('foo')`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,16 @@ The `get` method returns the item of the `ArrayElement` instance at the given in
 
 ```javascript
 var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
-var value = arrayElement.get(0) // get(0) returns 'a'
+var value = arrayElement.get(0) // get(0) returns item for 'a'
+```
+
+##### getValue
+
+The `getValue` method returns the value of the item of the `ArrayElement` instance at the given index.
+
+```javascript
+var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
+var value = arrayElement.getValue(0) // get(0) returns 'a'
 ```
 
 ##### set
@@ -363,12 +372,21 @@ This is an element for representing objects. Objects store their items as an ord
 
 ##### get
 
-The `get` method returns the value of the `ObjectElement` instance at the given name.
+The `get` method returns the `ObjectElement` instance at the given name.
 See `getKey` and `getMember` for ways to get more instances around a key-value pair.
 
 ```javascript
 var objectElement = new minim.ObjectElement({ foo: 'bar' });
 var value = objectElement.get('foo') // returns string instance for 'bar'
+```
+
+##### getValue
+
+The `getValue` method returns the value of the `ObjectElement` instance at the given name.
+
+```javascript
+var objectElement = new minim.ObjectElement({ foo: 'bar' });
+var value = objectElement.getValue('foo') // returns 'bar'
 ```
 
 ##### getKey
--- a/lib/base.js
+++ b/lib/base.js
@@ -113,85 +113,6 @@ var ElementRegistry = createClass({
 // Initiate a default Minim registry
 var registry = new ElementRegistry();
 
-var Meta = createClass({
-  constructor: function(meta) {
-    var self = this;
-    self.meta = meta || {};
-
-    _.keys(meta).forEach(function(key) {
-      self.meta[key] = registry.toElement(meta[key]);
-    });
-  },
-
-  toObject: function() {
-    var self = this;
-    var meta = {};
-
-    _.forEach(_.keys(self.meta), function(key) {
-      meta[key] = self.meta[key].toValue();
-    });
-
-    return meta;
-  },
-
-  getProperty: function(name, value) {
-    if (!this.meta[name]) {
-      this.meta[name] = registry.toElement(value);
-    }
-
-    return this.meta[name];
-  },
-
-  setProperty: function(name, value) {
-    this.meta[name] = registry.toElement(value);
-  }
-}, {}, {
-  id: {
-    get: function() {
-      return this.getProperty('id', '');
-    },
-    set: function(element) {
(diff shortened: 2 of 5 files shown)


smizell:
@danielgtaylor This will definitely need a look from you, as these are some big changes.

README.md
```diff
@@ -371,6 +380,15 @@ var objectElement = new minim.ObjectElement({ foo: 'bar' });
 var value = objectElement.get('foo') // returns string instance for 'bar'
 ```
 
+##### getValue
+
+The `getValue` method returns the value of the `ObjectElement` instance at the given name.
+
+```javascript
+var objectElement = new minim.ObjectElement({ foo: 'bar' });
+var value = objectElement.getValue('foo') // returns 'bar'
```
fosrias, line 389: seems just `objectElement.value(0)` would be sufficient. As a rubyist my pull is to keep `get` and `set` out of method names, so I am biased.

lib/base.js
```diff
@@ -237,14 +166,14 @@ BaseElement = createClass({
...
   convertAttributesToRefract: function(functionName) {
     var attributes = {};
-    var keys = _.keys(this.attributes);
+    var keys = this.attributes.keys();
 
     for (var i = 0; i < keys.length; i++) {
       var key = keys[i];
       if (this._attributeElementKeys.indexOf(key) !== -1) {
-        attributes[key] = this.attributes[key][functionName]();
+        attributes[key] = this.attributes.get(key)[functionName]();
```
fosrias, line 174: This is brittle if functionName does not exist. Should this have some guard?

lib/base.js
```diff
@@ -329,8 +258,32 @@ BaseElement = createClass({
     }
   },
 
-  // TODO: Mixin these mutator functions to share code with Meta class
-  // Subclassing causes circular redundancy
+  meta: {
+    get: function() {
+      if (!this._meta) {
+        this._meta = registry.toElement({});
```
fosrias, line 264: Shouldn't #toElement() just do this vs. having to pass in an object?

lib/base.js
```diff
@@ -656,8 +625,8 @@ var MemberElement = BaseElement.extend({
   toRefract: function() {
     return {
       element: this.element,
-      attributes: this.attributes,
-      meta: this.meta.toObject(),
+      attributes: this.attributes.toValue(),
```
fosrias: suggested: keep convention of specifying meta before attributes as you do elsewhere.

test/primitives-test.js
```diff
@@ -17,10 +17,10 @@ describe('Minim Primitives', function() {
       });
 
       it('should initialize the correct meta data', function() {
-        expect(el.meta.id.toValue()).to.equal('foobar');
```
fosrias, line 20: Shouldn't we still keep the verbose variant of the test to expose the actual API since the helper methods are just that. Or are we no longer publicly supporting this API?

test/subclass-test.js
```diff
@@ -54,18 +54,18 @@ describe('Minim subclasses', function() {
...
+      expect(myElement.attributes.get('foo').toValue()).to.be.a('string');
     });
   });
 
   describe('serializing attributes', function() {
     var myElement = new MyElement();
-    myElement.attributes.headers = new minim.ArrayElement(['application/json']);
-    myElement.attributes.headers.content[0].meta.name = 'Content-Type';
+    myElement.attributes.set('headers', new minim.ArrayElement(['application/json']));
+    myElement.attributes.get('headers').content[0].meta.set('name', 'Content-Type');
```
fosrias, line 68: Not the most self-documenting...

README.md
smizell, line 389: Added an issues to discuss this.

https://github.com/refractproject/minim/issues/49

lib/base.js
smizell, line 174: I've been wanting to discuss this code as well. Issue added https://github.com/refractproject/minim/issues/50

lib/base.js
smizell, line 264: `#toElement` takes a primitive value and converts it to an element. You can give it number, `true`, etc.

``` js
registry.toElement(0);
```

test/primitives-test.js
smizell, line 20: I think you're right.

lib/base.js
smizell: Fixed

test/primitives-test.js
smizell, line 20: Fixed

test/subclass-test.js
smizell, line 68: Will be addressed in #50 when we figure this out.

smizell:
@fosrias I've addressed all comments.

test/primitives-test.js
```diff
@@ -17,10 +17,10 @@ describe('Minim Primitives', function() {
...
-        expect(el.meta.class.toValue()).to.deep.equal(['a', 'b']);
-        expect(el.meta.title.toValue()).to.equal('Title');
-        expect(el.meta.description.toValue()).to.equal('Description');
+        expect(el.meta.get('id').toValue()).to.equal('foobar');
```
fosrias, line 20: So, I cannot do: `el.meta.id.toValue()` now? This seems extra verbose (shock me).

smizell, line 20: You can do `element.id`, but currently not `element.meta.id`. It's the same interface now for any object with helper methods on `element`.

fosrias:
@smizell Fine by me. Leave to @danielgtaylor to clarify anything.

zdne:
@fosrias @smizell 

>  Leave to danielgtaylor to clarify anything.

Just a heads up: As far as I know Daniel is for two weeks off so you may want to merge it without him (but I leave that up to you)

PR #53, opened 2015-06-20
Fix attribute bug

smizell:
While tinkering, found a bug where attributes were not converted correctly.


fosrias:
@smizell Looks good. Feel free to merge.

PR #55, opened 2015-06-23
Update functional methods

smizell:
This PR does the following:
- Updates how the `find` method works.
- Adds reduce function
- Adds clone function
- Updates `forEach` for Array Element to return indexes
- Updates `map` for Array and Object Elements to return Array Element

The last commit is a big change. It converts `map` to be a special kind of map, where instead of getting back a normal JavaScript array, you get back an Array Element. This follows with the idea that all of these instance methods are Minim-specific.

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,15 @@ var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
 var value = arrayElement.getValue(0) // get(0) returns 'a'
 ```
 
+##### getIndex
+
+The `getIndex` method returns the item of the array at a given index.
+
+```javascript
+var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
+var value = arrayElement.getIndex(0) // returns the item for 'a'
+```
+
 ##### set
 
 The `set` method sets the value of the `ArrayElement` instance.
@@ -267,6 +276,26 @@ var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
 var newArray = arrayElement.filter(function(item) {
   return item.get() === 'a'
 }); // newArray.toValue() is now ['a']
+```
+
+##### reduce
+
+The `reduce` method may be used to reduce over a Minim array or object. The method takes a function and an optional beginning value.
+
+```javascript
+var numbers = new minim.ArrayElement([1, 2, 3, 4]);
+var total = numbers.reduce(function(result, item) {
+  return result.toValue() + item.toValue();
+}); // total.toValue() === 10
+```
+
+The `reduce` method also takes an initial value, which can either be a value or Minim element.
+
+```javascript
+var numbers = new minim.ArrayElement([1, 2, 3, 4]);
+var total = numbers.reduce(function(result, item) {
+  return result.toValue() + item.toValue();
+}, 10); // total.toValue() === 20
 ```
 
 ##### forEach
--- a/README.md
+++ b/README.md
@@ -151,6 +151,15 @@ var stringElement = minim.convertToElement("foobar");
 stringElement.equals('abcd'); // returns false
 ```
 
+#### clone
+
+Creates a clone of the given instance.
+
+```javascript
+var stringElement = minim.convertToElement("foobar");
+var stringElementClone = stringElement.clone();
+```
+
 ### Minim Elements
 
 Minim supports the following primitive elements
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -125,19 +125,23 @@ module.exports = function(BaseElement, registry) {
       var recursive = !!options.recursive;
       var results = options.results === undefined ? [] : options.results;
 
-      this.content.forEach(function(el) {
+      // The forEach method for Object Elements returns value, key, and member.
+      // This passes those along to the condition function below.
+      this.forEach(function(item, keyOrIndex, member) {
         // We use duck-typing here to support any registered class that
         // may contain other elements.
-        if (recursive && (el.findElements !== undefined)) {
-          el.findElements(condition, {
+        if (recursive && (item.findElements !== undefined)) {
+          item.findElements(condition, {
             results: results,
             recursive: recursive
           });
         }
-        if (condition(el)) {
-          results.push(el);
+
+        if (condition(item, keyOrIndex, member)) {
+          results.push(item);
         }
       });
+
       return results;
     },
 
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -149,7 +149,9 @@ module.exports = function(BaseElement, registry) {
     },
 
     forEach: function(cb) {
-      this.content.forEach(cb);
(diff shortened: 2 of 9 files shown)


smizell:
cc: @fosrias

README.md
```diff
@@ -269,6 +287,17 @@ var newArray = arrayElement.filter(function(item) {
 }); // newArray.toValue() is now ['a']
 ```
 
+##### reduce
+
+The `reduce` method may be used to reduce over a Minim array or object. The method take a function and an optional beginning value.
```
fosrias: The method takes ...

README.md
```diff
@@ -269,6 +287,17 @@ var newArray = arrayElement.filter(function(item) {
...
+
+```javascript
+var numbers = new minim.ArrayElement([1, 2, 3, 4]);
```
fosrias, line 295: No ability to set a starting value for the collector?

lib/primitives/array-element.js
```diff
@@ -107,8 +116,47 @@ module.exports = function(BaseElement, registry) {
...
+          memo = registry.toElement(fn(memo, item));
+        }
+      }
+
+      return memo;
+    },
+
     forEach: function(cb) {
-      this.content.forEach(cb);
+      this.content.forEach(function(item, index) {
```
fosrias, line 152: Should this replace the item with the converted element version for future performance? Or does that already happen?

lib/primitives/base-element.js
```diff
@@ -17,7 +17,7 @@ module.exports = function(registry) {
         this.attributes.set(attributes);
       }
 
-      this.content = content || null;
+      this.content = content !== undefined ? content : null;
```
fosrias, line 20: Is there any difference here compared to original code?

lib/primitives/object-element.js
```diff
@@ -111,9 +111,11 @@ module.exports = function(BaseElement, ArrayElement, MemberElement) {
     },
 
     map: function(cb) {
```
fosrias: No collector?

lib/primitives/object-element.js
```diff
@@ -2,7 +2,7 @@
```
fosrias, line 2: No reduce?

test/primitives/array-element-test.js
```diff
@@ -133,10 +141,17 @@ describe('ArrayElement', function() {
...
       it('allows for mapping the content of the array', function() {
         var newArray = arrayElement.map(function(item) {
-          return item.toValue();
+          return item;
         });
-        expect(newArray).to.deep.equal(['a', true, null, 1]);
+        expect(newArray.toValue()).to.deep.equal(['a', true, null, 1]);
       });
+
+      it('converts primitive values', function() {
```
fosrias: This is a broad statement for using just integers in the test.

test/primitives/array-element-test.js
```diff
@@ -195,6 +236,15 @@ describe('ArrayElement', function() {
     //     expect(items).to.have.length(4);
     //   });
     // });
+
+    describe('#clone', function() {
+      it('creates a deep clone of the element', function() {
+        var clone = arrayElement.clone();
+        expect(clone).to.be.instanceOf(minim.ArrayElement);
+        expect(clone).to.not.equal(arrayElement);
```
fosrias, line 237: I think a clone should equal but not deep equal. Can that be done in javascript?

fosrias, line 237: Well, from the next expectation, I suppose I have these ideas switched in my mind in Javascript.

fosrias, line 237: Ah, well looking again you convert it for this to work. So is there no way for an element to == another but not deep equal (i.e. be the same instance)?

test/primitives/object-element-test.js
```diff
@@ -306,6 +313,40 @@ describe('ObjectElement', function() {
...
+
+    it('allows for reducing on keys', function() {
+      var letters = numbers.reduce(function(result, item, key) {
+        return result.push(key);
+      }, []);
+      expect(letters.toValue()).to.deep.equal(['a', 'b', 'c', 'd']);
+    });
+
+    context('when no beginning value is given', function() {
+      it('correctly reduces the array', function() {
```
fosrias: C&P error? There is no array

fosrias: ditto below

fosrias:
@smizell Main question on clone is about ability to define equality of objects in Javascript and so be able to compare them directly as equal if they have the same value but have different instance IDs I suspect this is not possible, but if it is, seems we should do that (Ruby leaking in to this opinion).

README.md
smizell, line 295: There is, but it's not mentioned here. Will add.

lib/primitives/base-element.js
smizell, line 20: This covers the issue where content could be falsy, and therefore gets `null`, though you may actually want it to be an empty string.

lib/primitives/array-element.js
smizell, line 152: Could you explain a little here? The `forEach` function is only for iterating, so it doesn't do anything with the value other than present it to the function as an argument. The function then handles the rest.

lib/primitives/object-element.js
smizell, line 2: It's inherited from ArrayElement

smizell:
@fosrias It is interesting to see things leak from both our sides. I think it provides helpful perspectives.

Here are some JavaScript samples.

```
> var a = { foo: 'bar' }
undefined
> a === { foo: 'bar' }
false
> var b = lodash.clone(a);
undefined
> a === b
false
> lodash.isEqual(a, b)
true
```

Minim actually wraps lodash's `isEqual`, so we get this deep equality, but it's only at the value level and not the Refract level, if that makes sense.

Clone converts the entire element to Refract serialization and then converts it back to element instances. This means their values and structures will be the same, but they are different instances.

With all of that said, I think clone works how it should work (which is how lodash's `cloneDeep` works) and will not provide the ability to do this checking you're wanting.

test/primitives/array-element-test.js
smizell, line 237: > be the same instance

Since these are objects, testing equality has to be handled at the value level and not based on the instance. This is because `{ foo: 'bar'} === { foo: 'bar'}` is `false`.

lib/primitives/array-element.js
fosrias, line 152: Nm.

lib/primitives/base-element.js
fosrias, line 20: Ok

danielgtaylor:
I think this looks pretty good. My only comment would be to add the tests from [my other PR](https://github.com/refractproject/minim/pull/64/files#diff-1cbcc6a779160f9828bdd574c2a75925R26) re: falsey values. :+1:

pksunkara:
Opened new PR #67

PR #56, opened 2015-06-27
Add functionality for embedded Refract

smizell:
This adds the functionality for embedded Refract. This is the initial commit of it, and I'm adding it mainly for exploring the format. Once I like it, I'll add it to the README.

cc: @fosrias

### DIFF
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -38,6 +38,18 @@ module.exports = function(BaseElement, registry) {
       return [this.element, this.meta.toValue(), attributes, compactDoms];
     },
 
+    toEmbeddedRefract: function() {
+      if (this.attributes.content.length === 0 && this.meta.content.length === 0) {
+        return this.map(function(item) {
+          return item.toEmbeddedRefract();
+        });
+      }
+
+      return {
+        refract: this.toRefract()
+      };
+    },
+
     fromRefract: function(doc) {
       this.meta = doc.meta;
       this.attributes = doc.attributes;
--- a/lib/primitives/base-element.js
+++ b/lib/primitives/base-element.js
@@ -45,6 +45,16 @@ module.exports = function(registry) {
       return [this.element, this.meta.toValue(), attributes, this.content];
     },
 
+    toEmbeddedRefract: function() {
+      if (this.attributes.content.length === 0 && this.meta.content.length === 0) {
+        return this.toValue();
+      }
+
+      return {
+        refract: this.toRefract()
+      };
+    },
+
     /*
      * Some attributes may be elements. This is domain-specific knowledge, so
      * a subclass *MUST* define the attribute element names to convert. This
@@ -111,6 +121,13 @@ module.exports = function(registry) {
       }
 
       return this;
+    },
+
+    fromEmbeddedRefract: function(embeddedObj) {
+      var self = this;
+      _.forEach(_.keys(embeddedObj), function(key) {
+        self.attributes.set(key, registry.fromEmbeddedRefract(embeddedObj[key]));
+      });
     },
 
     set: function(content) {
--- a/lib/primitives/object-element.js
+++ b/lib/primitives/object-element.js
@@ -2,7 +2,7 @@
 
 var _ = require('lodash');
 
-module.exports = function(BaseElement, ArrayElement, MemberElement) {
+module.exports = function(BaseElement, ArrayElement, MemberElement, registry) {
   var ObjectElement = ArrayElement.extend({
     constructor: function (content, meta, attributes) {
       var convertedContent = _.keys(content).map(function(key) {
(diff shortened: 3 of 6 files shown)


lib/primitives/array-element.js
```diff
@@ -38,6 +38,18 @@ module.exports = function(BaseElement, registry) {
       return [this.element, this.meta.toValue(), attributes, compactDoms];
     },
 
+    toEmbeddedRefract: function() {
+      if (this.attributes.content.length === 0 && this.meta.content.length === 0) {
```
fosrias, line 42: Seems like this could be abstracted into some helper method and make Demeter more happy.

lib/primitives/base-element.js
```diff
@@ -45,6 +45,16 @@ module.exports = function(registry) {
       return [this.element, this.meta.toValue(), attributes, this.content];
     },
 
+    toEmbeddedRefract: function() {
+      if (this.attributes.content.length === 0 && this.meta.content.length === 0) {
```
fosrias, line 49: Ditto and more DRY.

lib/primitives/object-element.js
```diff
@@ -23,6 +23,29 @@ module.exports = function(BaseElement, ArrayElement, MemberElement) {
       }, {});
     },
 
+    toEmbeddedRefract: function() {
+      if (this.attributes.content.length === 0 && this.meta.content.length === 0) {
```
fosrias, line 27: Ditto

lib/primitives/object-element.js
```diff
@@ -23,6 +23,29 @@ module.exports = function(BaseElement, ArrayElement, MemberElement) {
...
+
+        return refract;
+      }
+
+      return {
+        refract: this.toRefract()
+      };
+    },
+
+    fromEmbeddedRefract: function(embeddedObj) {
```
fosrias, line 42: This does not seem DRY. Can't this be inherited somehow?

smizell, line 42: Can you explain? The `fromEmbeddedRefract` is inherited, but this overwrites it since the Object Element acts a little different. On normal elements, the properties are stored as attributes, but for Object Elements, they are not since it is an actual object.

lib/primitives/array-element.js
smizell, line 42: Been wanting to add this for a while. Will do on a later PR.

https://github.com/refractproject/minim/issues/57

lib/primitives/base-element.js
smizell, line 49: https://github.com/refractproject/minim/issues/57

lib/primitives/object-element.js
smizell, line 27: https://github.com/refractproject/minim/issues/57

lib/primitives/object-element.js
fosrias, line 42: Looks the same as: https://github.com/refractproject/minim/pull/56/files#diff-223850cdafb14bcee834544fd3fc54e2R126

Am I wrong?

smizell, line 42: Yep, one is a setter on itself, another is setter on the attributes.

test/embedded-refract-test.js
```diff
@@ -0,0 +1,442 @@
...
+
+      it('returns the correct value', function() {
+        expect(element.toValue()).to.deep.equal(obj);
+      });
+    });
+
+    context('when given embedded refract object', function() {
+      // An object { b: 2 } with embedded refract
+      var obj = {
+        refract: {
```
fosrias, line 26: Hm. This seems a bit strange. So the idea is that if I take a non-refracted object, I can put a refract element in it that provides metadata about the object itself? Seems a bit of a mind-tweak.

test/embedded-refract-test.js
```diff
@@ -0,0 +1,442 @@
...
+      it('returns the correct value', function() {
+        expect(element.toRefract()).to.deep.equal(refract);
+      });
+    });
+
+    context('when a property is embedded refract', function() {
+      // Object { a: 1, b: 2, c: 3 } but the value of `b` is embedded refract
+      var obj = {
+        a: 1,
+        b: {
```
fosrias, line 84: What does the following mean:

``` javascript
b: {
      refract: {
        element: 'number',
        content: 2
    },
    d: 5
}
```

fosrias:
Suggestion for the future: Do this for smaller subset of Elements at a time. This is pretty big to review, particularly the tests.

fosrias:
I think this PR needs more justification or explanation of the value of it. At least in the main PR comment. Dropping into this I think I see the value, but this is a bit hard to review without more context.

test/embedded-refract-test.js
smizell, line 84: It means that `b` is a number with an attribute of `d` that is equal to 5.

smizell:
> I think this PR needs more justification or explanation of the value of it.

This was a weekend experiment, so I am fine with the lack of value defined. I think I would need to write up some spec on how embedded Refract would work to make this better understandable. Unfortunately, I think I will abandon this concept for at least the foreseeable future.

test/embedded-refract-test.js
fosrias, line 84: @smizell I don't think that is a consistent interpretation with what you see here: https://github.com/refractproject/minim/pull/56/files#diff-d4f2450b340a251071ab34ac7d28869eR25

smizell, line 84: That is because the other is an object. Because of that, it treats the object as an actual object where the properties are actual properties of the object. 

These other tests cover, what happens when you get a property on a non-object element? We store that data in attributes to preserve that data.

PR #58, opened 2015-06-28
Events

smizell:
Note: this is experimental at this point. Wanting to see what it would look like to be able to listen to changes at different nodes in the tree.

### DIFF
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -94,7 +94,7 @@ module.exports = function(BaseElement, registry) {
 
     set: function(index, value) {
       var item = registry.toElement(value);
-      if (_.indexOf(item.parents, this) === -1) {
+      if (!_.contains(item.parents, this)) {
         item.parents.push(this);
       }
       this.content[index] = item;
@@ -117,7 +117,7 @@ module.exports = function(BaseElement, registry) {
 
     push: function(value) {
       var item = registry.toElement(value);
-      if (_.indexOf(item.parents, this) === -1) {
+      if (!_.contains(item.parents, this)) {
         item.parents.push(this);
       }
       this.content.push(item);
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -5,8 +5,11 @@ var _ = require('lodash');
 module.exports = function(BaseElement, registry) {
   var ArrayElement = BaseElement.extend({
     constructor: function (content, meta, attributes) {
+      var self = this;
       var convertedContent = (content || []).map(function(value) {
-        return registry.toElement(value);
+        var item = registry.toElement(value);
+        item.parents.push(self);
+        return item;
       });
       BaseElement.call(this, convertedContent, meta || {}, attributes || {});
       this.element = 'array';
@@ -98,6 +101,7 @@ module.exports = function(BaseElement, registry) {
         item.parents.push(this);
       }
       this.content[index] = item;
+      this.emitSwap();
       return this;
     },
 
@@ -121,6 +125,7 @@ module.exports = function(BaseElement, registry) {
         item.parents.push(this);
       }
       this.content.push(item);
+      this.emitSwap();
       return this;
     },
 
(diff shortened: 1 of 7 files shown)


lib/primitives/array-element.js
```diff
@@ -93,7 +96,12 @@ module.exports = function(BaseElement, registry) {
     },
 
     set: function(index, value) {
-      this.content[index] = registry.toElement(value);
+      var item = registry.toElement(value);
+      if (!_.contains(item.parents, this)) {
+        item.parents.push(this);
+      }
+      this.content[index] = item;
+      this.emitSwap();
```
fosrias, line 104: swap? Sort of confusing semantics. This is replacing the element at some position with another, is that correct?

lib/primitives/base-element.js
```diff
@@ -132,6 +135,19 @@ module.exports = function(registry) {
 
     setMetaProperty: function(name, value) {
       this.meta.set(name, value);
+    },
+
+    emitSwap: function() {
```
fosrias, line 140: This does not seem to correlate the objects related to a particular event as to what was replaced by what if I understand this correctly. Seems like something is missing with respect to the event. Should this have some event object that contains the information?

test/primitives/array-element-test.js
```diff
@@ -128,6 +128,26 @@ describe('ArrayElement', function() {
...
+      });
+
+      it('triggers swap on the parents', function(done) {
+        var arr = new minim.ArrayElement([1]);
+        arr.on('swap', done);
+        arr.set(1, 2);
+      });
+
+      context('when a value is changed directly', function() {
+        it('triggers swap on the parents', function(done) {
```
fosrias, line 145: Is this "bubbling" the event, or actually a separate event?

test/primitives/array-element-test.js
```diff
@@ -176,6 +196,18 @@ describe('ArrayElement', function() {
         arrayElement.push('foobar');
         itAddsToArray(arrayElement);
       });
+
+      it('sets the parent to children', function() {
+        var item = new minim.StringElement('foobar');
+        arrayElement.push(item);
+        expect(item.parents).to.include(arrayElement);
```
fosrias, line 203: Seems this functionality should be private and associated only with event functionality if that is really the purpose of parents here. Not clear.

test/primitives/object-element-test.js
```diff
@@ -207,6 +207,47 @@ describe('ObjectElement', function() {
       obj.set({ foo: 'bar' });
       expect(obj.get('foo').toValue()).to.equal('bar');
     });
+
+    it('sets the parent to children', function() {
```
fosrias, line 211: Are we normally tracking parents or why are we tracking parents?

lib/primitives/array-element.js
smizell, line 104: Swap is a term used for when new instances and memory allocations are changed in atomic structures. Swap means that a node on the tree has changed. See description here http://omniscientjs.github.io/. 

Note: I want to add more events that can be listened to, such as `update` or `add`.

lib/primitives/base-element.js
smizell, line 140: I would handle passing the changed value in another event, like `update`. This is merely saying the branch of the tree has changed.

test/primitives/array-element-test.js
smizell, line 145: Each event is on a separate node. So when you trigger an event on a node, it triggers the swap events all of the way up. This means you can listen for changes on specific parts of the tree.

test/primitives/object-element-test.js
smizell, line 211: Parents let you go up and down the tree. Right now, if you get a node from the tree, you cannot know where it is in the tree, only what its children are.

test/primitives/array-element-test.js
smizell, line 203: Parent just provides access to the node's parents. This is useful in functions like `find` where you want to find certain elements that have a specific kind of parent. Also allows traversing up the tree.

kylef:
Closing, this was solved by #141.

PR #60, opened 2015-06-28
Add getIn method for array elements

smizell:
Given a tree like this:

``` js
{
  a: {
    b: {
      c: [1, 2, 3]
    }
  }
}
```

You can currently do this:

``` js
obj
  .get('a')
  .get('b')
  .get('c')
  .get(1)
```

This would give you the value of 2.

With `getIn`, you can now do.

``` js
obj.getIn(['a', 'b', 'c', 1]);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,18 @@ var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
 var value = arrayElement.getValue(0) // get(0) returns 'a'
 ```
 
+##### getIn
+
+The `getIn` method returns the value of the item at the given path.
+
+```javascript
+var arrayElement = new minim.ArrayElement([{
+  a: [1, 2, 3]
+}]);
+var value = arrayElement.getIn([0, 'a', 1]) // returns the element for 2
+```
+
+
 ##### set
 
 The `set` method sets the value of the `ArrayElement` instance.
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -92,6 +92,15 @@ module.exports = function(BaseElement, registry) {
       return undefined;
     },
 
+    /*
+     * It reduces through a given array and calls get on each item.
+     */
+    getIn: function(paths) {
+      return _.reduce(paths, function(result, path) {
+        return result.get(path);
+      }, this);
+    },
+
     set: function(index, value) {
       this.content[index] = registry.toElement(value);
       return this;
--- a/test/primitives/array-element-test.js
+++ b/test/primitives/array-element-test.js
@@ -123,6 +123,16 @@ describe('ArrayElement', function() {
       });
     });
 
+    describe('#getIn', function() {
+      var arr = new minim.ArrayElement([
+        {
+          a: [1, 2, 3]
+        }
+      ]);
+
+      expect(arr.getIn([0, 'a', 1]).toValue()).to.equal(2);
+    });
+
     describe('#set', function() {
       it('sets the value of the array', function() {
         arrayElement.set(0, 'hello world');


README.md
```diff
@@ -237,6 +237,18 @@ var arrayElement = new minim.ArrayElement(['a', 'b', 'c']);
...
 
+##### getIn
+
+The `getIn` method returns the value of the item at the given path.
+
+```javascript
+var arrayElement = new minim.ArrayElement([{
+  a: [1, 2, 3]
+}]);
+var value = arrayElement.getIn([0, 'a', 1]) // returns the element for 2
```
fosrias, line 248: Suggested: returns the element describing the number 2.

lib/primitives/array-element.js
```diff
@@ -92,6 +92,15 @@ module.exports = function(BaseElement, registry) {
       return undefined;
     },
 
+    /*
+     * It reduces through a given array and calls get on each item.
+     */
+    getIn: function(paths) {
```
fosrias, line 98: Is #getIn the best name? And your description is even more implementation specific vs. function specific here. Something to think about is a more representative name for the function and documentation that describes the end result vs how it does it.

test/primitives/array-element-test.js
```diff
@@ -123,6 +123,16 @@ describe('ArrayElement', function() {
...
     });
 
+    describe('#getIn', function() {
+      var arr = new minim.ArrayElement([
+        {
+          a: [1, 2, 3]
+        }
+      ]);
+
+      expect(arr.getIn([0, 'a', 1]).toValue()).to.equal(2);
```
fosrias, line 133: So, just a comment on these tests that they sort of hide the interface. Assuming that elements have equality defined, I would find the following more ilumintating and self documenting:

``` javascript

describe('#getIn', function() {
      var arrayElement = new minim.ArrayElement([
        {
          a: [1, 2, 3]
        }
      ]);

      nested_element = new Element(2);
      expect(arrayElement.getIn([0, 'a', 1])).to.equal(nested_element);
    });
```

Calling toValue() all over the place obscures the interface and is less articulate IMO.

lib/primitives/array-element.js
smizell, line 98: The name `getIn` is more from the functional programming world, specifically [Clojure](https://clojuredocs.org/clojure.core/get-in). Also see [Mori.js](http://swannodette.github.io/mori/#getIn). Will think about different name.

test/primitives/array-element-test.js
smizell, line 133: This test will not work, as equality will not be met since they are different instances of and object. It's like when you say:

``` js
expect({ foo: 'bar'}).to.equal({ foo: 'bar'})
```

and you get a failing test. You have to use deep equal for that situation to test against the actual values given.

So I'm not actually sure the of the way to do this without converting back to a value and comparing against primitive values. I could compare against the content of the value, but I am trying to stay away from accessing content directly. I'm open to other ideas, but I'm afraid `.toValue` or `.toRefract` may be the best option.

I do agree with your issue here though.

lib/primitives/array-element.js
smizell, line 98: And, any thoughts on better name? I can overload the `#get` method, but I'm not sure I like that.

fosrias, line 98: I could go with over loading #get to accept paths to something. You are getting an element, but just navigating a path to arrive there.

smizell, line 98: This makes for a pretty complicated interface, IMO. You're OK with adding in this kind of magic? 

If we decide to do this, I think I will leave `getIn` and just delegate this functionality to it in `get`.

smizell:
@kylef can I propose to close this and let you all address whether to include this idea or not? Seems like it's not necessary at the moment.

kylef:
I don't think its necessarily, but I am not sure that I understand the use case.

> You can currently do this:
> 
> ``` js
> obj
>   .get('a')
>   .get('b')
>   .get('c')
>   .get(1)
> ```

Perhaps it would make more sense to search for the specific element you want and get the value to prevent unnecessary coupling of the tree.

Example:

```js
const method = element.recursiveChildren.find('httpRequest').method
```

Instead of

```js
const method = element.get(0).get(0).get(0).method;
```

kylef:
I'm going to close this like suggested as it has been open for a while and underlying code will no longer merge.

PR #63, opened 2015-07-10
Ref Element

smizell:
Provide partial support for the [Ref element](https://github.com/refractproject/refract-spec/blob/master/refract-spec.md#referencing-and-linking).


kylef:
Handled in #119.

PR #64, opened 2015-07-20
Properly handle falsey element content values.

danielgtaylor:
This fixes the following issue:

``` js
let el = new minim.StringElement('');
expect(el.toValue()).to.equal('');
// => Fails with `null` !== `''`
```

The same thing happens for numbers and boolean elements.

---

On a related note, we may want to consider using `undefined` for the content instead of `null` to denote 'nothing is set' on the element. That makes this code even simpler by just using `this.content = content`. I forget if there is a reason to use `null`... thoughts?

cc @smizell

### DIFF
--- a/lib/primitives/base-element.js
+++ b/lib/primitives/base-element.js
@@ -17,7 +17,7 @@ module.exports = function(registry) {
         this.attributes.set(attributes);
       }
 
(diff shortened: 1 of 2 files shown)


danielgtaylor:
Turns out this is already done with #55.

PR #66, opened 2015-07-21
Return value result from simple types for convenience attributes.

danielgtaylor:
This changes the return value of the convenience attributes (e.g. `id`,
`title`, etc) to return the result of `.toValue()`. It **only** does so for
simple types (string, number, boolean, null) and leaves more complex types
alone.

``` js
console.log(element.title); // => 'I am a title'
console.log(element.description); // => ''

// Since `class` is an array, it is still an element instance
if (element.class.contains('foo')) {
  console.log('Contains foo.');
}
```

cc @smizell - any other thoughts on whether `.class` should be an array vs. ArrayElement instance? It seems super useful to keep it as an element instance, but makes accessing the strings within a bit harder.

### DIFF
--- a/test/primitives/base-element-test.js
+++ b/test/primitives/base-element-test.js
@@ -84,11 +84,11 @@ describe('BaseElement', function() {
     });
 
     it('returns true when they are equal', function() {
-      expect(el.id.equals('foobar')).to.be.true;
+      expect(el.meta.get('id').equals('foobar')).to.be.true;
     });
 
     it('returns false when they are not equal', function() {
-      expect(el.id.equals('not-equal')).to.be.false;
+      expect(el.meta.get('id').equals('not-equal')).to.be.false;
     });
 
     it('does a deep equality check', function() {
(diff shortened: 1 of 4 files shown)


test/primitives/base-element-test.js
```diff
@@ -84,11 +84,11 @@ describe('BaseElement', function() {
     });
 
     it('returns true when they are equal', function() {
-      expect(el.id.equals('foobar')).to.be.true;
```
pksunkara, line 87: So, basically, we have decided that `el.id` giving the actual value is more common than this, right?

danielgtaylor, line 87: @pksunkara yes, this allows this sort of (very common) stuff:

``` js
// Simple value access
console.log(el.id);

// Simple value assignment
el.id = 'abc123';

// You still have full access to the element the long way:
el.meta.get('id').attributes.set('encoding', 'utf-8');
```

That last example isn't realistic for the ID, but a language or encoding attribute might make sense for description or response body content in the API namespace.

pksunkara:
I guess, we need to release a minor version after the PR instead of a patch.

PR #67, opened 2015-07-27
Update functional methods

danielgtaylor:
This includes and supersedes #55. It includes the tests from #64. It is required for apiaryio/fury.js#38 Should be ready to merge.

cc @smizell

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,30 @@ var total = numbers.reduce(function(result, item) {
 }, 10); // total.toValue() === 20
 ```
 
+The `reduce` method also works with objects:
+
+```javascript
+var objNumbers = new minim.ObjectElement({a: 1, b:2, c:3, d:4});
+var total = objNumbers.reduce(function(result, item) {
+  return result.toValue() + item.toValue();
+}, 10); // total.toValue() === 20
+```
+
+The function passed to `reduce` can accept up to five optional parameters and depends on whether you are using an array element or object element:
+
+**Array**
+1. `result`: the reduced value thus far
+2. `item`: the current item in the array
+3. `index`: the zero-based index of the current item in the array
+4. `arrayElement`: the array element which contains `item` (e.g. `numbers` above)
+
+**Object**
+1. `result`: the reduced value thus far
+2. `item`: the value element of the current item in the object
+3. `key`: the key element of the current item in the object
+4. `memberElement`: the member element which contains `key` and `value`
+5. `objectElement`: the object element which contains `memberElement` (e.g. `objNumbers` above)
+
 ##### forEach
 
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -269,6 +287,26 @@ var newArray = arrayElement.filter(function(item) {
...
+```
+
+The `reduce` method also takes an initial value, which can either be a value or Minim element.
+
+```javascript
+var numbers = new minim.ArrayElement([1, 2, 3, 4]);
+var total = numbers.reduce(function(result, item) {
+  return result.toValue() + item.toValue();
+}, 10); // total.toValue() === 20
+```
```
pksunkara, line 308: I think we should show an example on how to reduce over an `ObjectElement`

lib/primitives/array-element.js
```diff
@@ -107,8 +111,47 @@ module.exports = function(BaseElement, registry) {
...
+      // Sending each function call to the registry allows for passing Minim
+      // instances through the function return. This means you can return
+      // primitive values or return Minim instances and reduce will still work.
+      for (var i = startIndex; i < this.length; i++) {
+        var item = this.content[i];
+
+        if (this.primitive() === 'object') {
+          memo = registry.toElement(fn(memo, item.value, item.key, item));
+        } else {
+          memo = registry.toElement(fn(memo, item));
```
pksunkara: We might want to use `fn(memo, item, i, this.content)` to mimic the [`Array.prototype.reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce)

README.md
pksunkara, line 308: I am asking this because not many know the standard for the parameters sent to the callback function for an `object.reduce`. (I don't even know if there is an object.reduce in any language).

pksunkara:
Review done.

lib/primitives/array-element.js
smizell: That's a good idea, IMO.

smizell:
@danielgtaylor I'm good with this :) I'll leave the addressing of comments by @pksunkara up to you. Thanks for the extra tests.

README.md
danielgtaylor, line 308: I've added an example in the latest commit.

lib/primitives/array-element.js
danielgtaylor: @pksunkara @smizell please have another look at the latest commit. I believe that exposing the element instance rather than the underlying array of `this.content` is better because it will give us access to any metadata and attributes defined on the array. The same goes for how I've modified the object reduce. I've also added some tests to ensure these values are always passed to the function like we expect.

PR #71, opened 2015-07-29
Run tests in the browser as well

XVincentX:
This small PR will ensure that all current tests will be run in a browser instance using `karma`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,3 +3,6 @@ node_js:
   - "iojs"
   - "0.12"
   - "0.10"
+before_install:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -34,7 +34,7 @@ module.exports = function(config) {
     // test results reporter to use
     // possible values: 'dots', 'progress'
     // available reporters: https://npmjs.org/browse/keyword/karma-reporter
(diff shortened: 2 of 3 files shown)


pksunkara:
:scream:

Small question. Why do we need to test this since we are not using this library in the browser?

XVincentX:
@pksunkara 
Basically because we will need minim in the browser very soon. We will use it to render MSON attributes in the documentation with the new AST.

smizell:
Ping @danielgtaylor

danielgtaylor:
@XVincentX @pksunkara @smizell what do you think of only running the frontend tests on the CI hosts (or using phantomjs locally)? Having a Firefox window pop up every time that I run `npm test` is a bit jarring.

pksunkara:
I agree with you.

smizell:
@danielgtaylor There is a command for running just Node tests, added by @XVincentX in this request. This means that `npm test` runs all of the tests, while other commands are provided to do subsets of tests (such as `npm nodejstests`). 

I'm fine making frontend tests run on CI only if that would be best for everyone.

danielgtaylor:
Yeah I think this is probably fine. I'll just run `npm run nodejstests` for now and if it gets to be a problem I'll submit a PR to make the frontend tests only run on the CI hosts.

smizell:
:+1:

PR #72, opened 2015-08-14
Implement RFC 0003. Renames class -> classes

danielgtaylor:
Here is the [RFC](https://github.com/refractproject/rfcs/blob/master/text/0003-class-rename.md).

Note that this will break downstream users (e.g. Fury.js).

cc @smizell


smizell:
@danielgtaylor This looks good to me!

Should we publish a new version? In another PR?

pksunkara:
Please make a PR for release.

PR #73, opened 2015-08-16
Fix registry

smizell:
With the current implementation in master, there is no way to have a custom registry. This is because we initialize a registry and then every primitive element closures around it. When this happens, any time a new custom registry is used, the original elements will not know about that new registry.

To fix this, we have to generate new elements for each registry. It also requires changing the interface to expose the new registry.

One other thing we could do would be to have `minim.common` or `minim.global` as a kind of global registry that simply calls `init()` on require (and is subsequently cached).

@danielgtaylor Let me know what you think. Sorry about this interface change, but I think it's necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ var el = {
 ### Converting Javascript Values into Elements
 
 ```javascript
-var minim = require('minim');
+var minim = require('minim').init();
 var arrayElement = minim.convertToElement([1, 2, 3]);
 var refract = arrayElement.toRefract();
 ```
@@ -540,7 +540,7 @@ const values = objectElement.map((value, key, member) => {
 Minim allows you to register custom elements. For example, if the element name you wish to handle is called `category` and it should be handled like an array:
 
 ```javascript
-var minim = require('minim');
+var minim = require('minim').init();
 
 // Register your custom element
 minim.registry.register('category', minim.ArrayElement);
@@ -554,6 +554,14 @@ console.log(elements.get(0).content); // hello, world
 
 // Unregister your custom element
 minim.registry.unregister('category');
+```
+
+You may also pass in a registries that may have been initialized elsewhere.
+
+```javascript
+var ElementRegistry = require('minim').ElementRegistry;
+var registry = new ElementRegistry();
+var minim = require('minim').init(registry);
 ```
 
 ### Chaining
--- a/lib/base.js
+++ b/lib/base.js
@@ -1,113 +1,10 @@
 'use strict';
 
+var _ = require('lodash');
+
 var uptown = require('uptown');
 var createClass = uptown.createClass;
-
-var BaseElement;
-
-/*
- * A refract element implementation with an extensible element registry.
- *
- * The element registry allows you to register your own classes to be instantiated
- * when a particular refract element is encountered, and allows you to specify
- * which elements get instantiated for existing Javascript objects.
- */
-
-var ElementRegistry = createClass({
-  constructor: function() {
-    this.elementMap = {};
-    this.elementDetection = [];
-  },
-
-  /*
-   * Register a new element class for an element.
-   */
-  register: function(name, ElementClass) {
-    this.elementMap[name] = ElementClass;
-    return this;
-  },
-
-  /*
-   * Unregister a previously registered class for an element.
-   */
-  unregister: function(name) {
-    delete this.elementMap[name];
-    return this;
-  },
-
-  /*
-   * Add a new detection function to determine which element
-   * class to use when converting existing js instances into
-   * refract element.
-   */
-  detect: function(test, ElementClass, givenPrepend) {
-    var prepend = givenPrepend === undefined ? true : givenPrepend;
-
-    if (prepend) {
-      this.elementDetection.unshift([test, ElementClass]);
-    } else {
-      this.elementDetection.push([test, ElementClass]);
-    }
-
-    return this;
-  },
-
-  /*
-   * Convert an existing Javascript object into refract element instances, which
-   * can be further processed or serialized into refract or compact refract.
-   * If the item passed in is already refracted, then it is returned
-   * unmodified.
-   */
-  toElement: function(value) {
-    if (value instanceof BaseElement) { return value; }
-
-    var element;
-
-    for(var i = 0; i < this.elementDetection.length; i++) {
-      var test = this.elementDetection[i][0];
-      var ElementClass = this.elementDetection[i][1];
-
-      if (test(value)) {
-        element = new ElementClass(value);
-        break;
-      }
-    }
-
-    return element;
-  },
-
-  /*
-   * Get an element class given an element name.
-   */
-  getElementClass: function(element) {
-    var ElementClass = this.elementMap[element];
-
-    if (ElementClass === undefined) {
-      // Fall back to the base element. We may not know what
-      // to do with the `content`, but downstream software
-      // may know.
-      return BaseElement;
-    }
-
-    return ElementClass;
-  },
-
-  /*
-   * Convert a long-form refract document into refract element instances.
-   */
-   fromRefract: function(doc) {
-    var ElementClass = this.getElementClass(doc.element);
-    return new ElementClass().fromRefract(doc);
-  },
-
-  /*
-   * Convert a compact refract tuple into refract element instances.
-   */
-  fromCompactRefract: function(tuple) {
-    var ElementClass = this.getElementClass(tuple[0]);
-    return new ElementClass().fromCompactRefract(tuple);
-  }
-});
+var ElementRegistry = require('./registry').ElementRegistry;
 
 // Initiate a default Minim registry
 var registry = new ElementRegistry();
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -39,7 +39,7 @@ var el = {
 ### Converting Javascript Values into Elements
 
 ```javascript
-var minim = require('minim');
+var minim = require('minim').init();
```
danielgtaylor, line 42: I would love to see this change go away. One way to do that would be to expose a default registry like we do today that just calls `init` internally. We should think about how to make this as simple as possible and I think having the minim module be a default registry instance might be an interesting idea. I actually do [that with Fury](https://github.com/apiaryio/fury.js/blob/master/src/fury.es6#L99-L112).

lib/base.js
```diff
@@ -1,138 +1,69 @@
...
-    return new ElementClass().fromRefract(doc);
-  },
-
-  /*
-   * Convert a compact refract tuple into refract element instances.
-   */
-  fromCompactRefract: function(tuple) {
-    var ElementClass = this.getElementClass(tuple[0]);
-    return new ElementClass().fromCompactRefract(tuple);
+exports.init = function(registry) {
```
danielgtaylor, line 7: I understand the use of `init` here but in many node.js modules they just export a default function. Also and perhaps more important, we should discuss how we plan to load the namespaces. Many, many node packages have a `.use(require('minim-api-description'))` or something like that. It's chainable and lets you load a bunch of plugins.

lib/base.js
```diff
@@ -1,138 +1,69 @@
...
+  // The registry needs a base element to determine the base type of an element.
+  registry.BaseElement = BaseElement;
+
+  var NullElement = require('./primitives/null-element')(BaseElement);
+  var StringElement = require('./primitives/string-element')(BaseElement);
+  var NumberElement = require('./primitives/number-element')(BaseElement);
+  var BooleanElement = require('./primitives/boolean-element')(BaseElement);
+  var ArrayElement = require('./primitives/array-element')(BaseElement, registry);
+  var MemberElement = require('./primitives/member-element')(BaseElement, registry);
+  var ObjectElement = require('./primitives/object-element')(BaseElement, ArrayElement, MemberElement);
```
danielgtaylor, line 27: I regularly subclass these in Fury (which will get split into it's own module soon). How does this work in the new system?

lib/base.js
```diff
@@ -1,138 +1,69 @@
...
+    registry: registry,
+    convertToElement: function() {
+      return registry.toElement.apply(registry, arguments);
+    },
+    convertFromRefract: function() {
+      return registry.fromRefract.apply(registry, arguments);
+    },
+    convertFromCompactRefract: function() {
+      return registry.fromCompactRefract.apply(registry, arguments);
+    }
```
danielgtaylor, line 67: This object returned from `init` seems strange. It seems like a module but it's not really. I feel a bit weird about this, but don't have much constructive input :smile:

smizell, line 67: Very true, but I can't figure out a way around it in my head. If you want to be able to have your own registry apart from the global one (which this PR does not have at the moment), you have to initialize all elements with that registry. It's the circular dependency issue. 

For example, the ArrayElement needs to have the registry to convert between types, but if you export it as a module on require, it will always have that one registry. Creating a custom registry will never work because ArrayElement doesn't know about it unless you store the new registry for each instance.

Storing a registry on each instance is a fundamental change, too, but may actually be the correct one given namespaces. But this still requires having some class that manages the scope (i.e. available registry) for a given document.

README.md
smizell, line 42: Couple of questions:
1. Can you have this default export without ES6?
2. If not, and if you make the minim default a new registry, how can you expose the functions for defining your own registries?

lib/base.js
smizell, line 7: > they just export a default function

:+1: sounds good to me

> many node packages have a `.use(require('minim-api-description'))` or something like that. It's chainable and lets you load a bunch of plugins

I like this as well.

lib/base.js
smizell, line 27: You will have to initialize some registry OR use something that actually mimics the above, where you wrap your classes in a function that receive an initialized registry and elements.

smizell, line 27: This would be made possible per what you've suggested above.

https://github.com/refractproject/minim/pull/73#discussion_r37405835

README.md
danielgtaylor, line 42: 1. Yes. For example:
   
   ``` js
   var registry = new ElementRegistry();
   module.exports = registry;
   ```
2. Either we expose `ElementRegistry` as a property of itself or we do that as a special case with the default export. E.g:
   
   ``` js
   ElementRegistry.prototype.ElementRegistry = ElementRegistry;
   ```
   
   or
   
   ``` js
   registry.ElementRegistry = ElementRegistry;
   module.exports = registry;
   ```

Either way you'll be able to:

``` js
var minim = require('minim');
var data;

// Use default registry
data = minim.toElements(['abc', 123]);

// Use a custom registry
registry = new minim.ElementRegistry();
data = registry.toElements(['abc', 123]);
```

Note, this requires moving some methods we have floating around at the module level today into the element registry. I think this makes sense though.

I also think every registry should just default to loading the elements included with Minim. The namespace loading mechanism should only apply to namespaces that aren't the base on top of which everything else is built.

smizell, line 42: > I also think every registry should just default to loading the elements included with Minim

The problem is, if you don't do what I've done here (or actually store the registry in each element instance), you MUST always use the default registry. That is because as we currently have it, each instance only knows of one registry, and even if you create your own, those instances will only use the default. 

So the problem isn't really loading the including elements, it's that those elements (in master) only know of one registry.

If that makes sense.

danielgtaylor:
The interface I'd like to see:

``` js
var minim = require('minim');
var ElementRegistry = minim.ElementRegistry;
var data;

// Use the default registry to convert to/from refract elements
data = minim.toElement(['abc', 123]);
data = minim.fromRefract({element: 'string', content: 'foo'});
data = minim.fromCompactRefract(['string', {}, {}, 'foo']);

// Load namespaces into the default registry
minim.use(require('minim-data-structures'))
     .use(require('minim-api-description'))
     .use(require('minim-parse-result'));

// Now we can load a parse result with API elements!
data = minim.fromRefract(require('./parse-result.json'));
console.log(data.api.title); // => 'My API'

// Using a custom registry
registry = new ElementRegistry();
// The registry is now loaded with the default element set, but we can add more if we like:
registry.use(require('minim-data-structures'))
        .use(require('my-custom-namespace'));

data = minim.fromRefract(require('./custom-result.json');
console.log(data.custom.title); // => 'Custom test'

// Instantiate a single element by its class
var StringElement = registry.getElementClass('string');
console.log((new StringElement('foo')).toValue()); // => 'foo'

// This also works for subclassing, e.g. your namespace module might have:
exports.init = function (registry) {
  var StringElement = registry.getElementClass('string');

  var CustomElement = createClass(StringElement, {
    // define you custom class here...
  });

  registry.register('custom', CustomElement);
}

// Then the `.use(...)` mechanism above is really simple to implement:
ElementRegistry.prototype.use = function (mod) {
  mod.init(this);
}
```

Hope this makes sense. Sorry for the brain dump!

smizell:
@danielgtaylor While I like this, I unfortunately think, without doing something like I've done or other fancy tricks, that your proposal above will not work when you get to a custom registry. I'll use your example.

First, how things currently work:
1. A registry is [initialized](https://github.com/refractproject/minim/blob/041db69dcc56eb1d58dc510594684c0f0358da03/lib/base.js#L113) in `base.coffee`
2. Elements are [loaded](https://github.com/refractproject/minim/blob/041db69dcc56eb1d58dc510594684c0f0358da03/lib/base.js#L118-L125) and passed this registry
3. Each element method that uses this registry [closures around the registry](https://github.com/refractproject/minim/blob/041db69dcc56eb1d58dc510594684c0f0358da03/lib/primitives/array-element.js#L9) from `base.coffee`

So now, even if I create a new registry as you've done, when I do:

``` js
registry.register('custom', CustomElement);
```

The original Minim elements will have no visibility of this new element, even though they are included in the new registry! Of course, if you register this custom element in the global registry it will work, but here with a custom registry, it won't.

I have a thought for a proposal, but I'd like to make sure I'm right here (I _think_ I am but that usually doesn't mean much).

danielgtaylor:
@smizell any more progress made on this? Here is my proposal, hopefully written out to be a bit easier to understand:

https://gist.github.com/danielgtaylor/fe17d92c3e0679fb8ac9

It's very similar to what you've proposed but changes the public interface slightly to match up with Node.js norms and be a bit simpler (in my biased opinion). How it works:
1. The minim module's default export is just an instance of an `ElementRegistry` that has been loaded with the default elements. It _also_ exposes the `ElementRegistry` class so you can use custom registries.
2. Namespaces are loaded via the registry's `use` method, which in turn just calls an `init` function in the namespace module. This is just like tons of other Node.js modules.
3. By default, no elements are loaded in a new registry. A `useDefault` method is added to load the default elements. This means one more step when using custom registries but allows you to build the registry entirely from scratch. **Alternative idea**: auto-load defaults and allow a method to clear the registry? Maybe make this a constructor option instead, but what about loading it later?
4. When creating a new namespace of elements, you must use the `registry.getElementClass` method to fetch previously defined elements. This ensures that you get the right element for the given registry, particularly if previously loaded namespaces may have replaced the default elements (is this safe and reasonable?) You must also define your element classes within the registry closure, just like you have done for Minim already.

Let me know if this isn't clear or you still think it wouldn't work!

danielgtaylor:
@smizell I've done a sample implementation in #77 to show how it would all work together. Let me know what you think. This is blocking me from splitting code out of Fury, which I'd like to get to within the next couple of weeks if possible.

smizell:
@danielgtaylor I like what you've done with `use`. I think when you create a new registry, it should have the defaults there with a method to remove them. Without them, it's just a registry of some sort (which might actually be a pattern we could abstract, but not now). 

I've added some comments on #77. Would be happy to carry on the conversation there so we aren't bouncing back and forth :)

PR #75, opened 2015-08-20
Add a toValue() method to member elements

danielgtaylor:
This is useful for elements that aren't objects but that use members, for example HTTP headers in the API Description namespace. The element is an array of member elements, and when converted to a js value it would be nice to get out an array of tuples.

cc @smizell

### DIFF
--- a/lib/primitives/member-element.js
+++ b/lib/primitives/member-element.js
@@ -12,6 +12,10 @@ module.exports = function(BaseElement, registry) {
       this.element = 'member';
     },
 
+    toValue: function () {
+      return [this.key.toValue(), this.value.toValue()];
+    },
+
     toRefract: function() {
       return {
         element: this.element,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+- Add a `.toValue()` method to member elements which returns a tuple of the
+  key and value.
+
 # 0.10.0 - 2015-08-18
 
 - Rename the `class` metadata property to `classes`. The convenience property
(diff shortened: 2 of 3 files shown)


lib/primitives/member-element.js
```diff
@@ -12,6 +12,10 @@ module.exports = function(BaseElement, registry) {
       this.element = 'member';
     },
 
+    toValue: function () {
+      return [this.key.toValue(), this.value.toValue()];
+    },
```
smizell, line 17: Question: do you want to return this array or would you rather have a hash?

``` js
return {
  key: this.key.toValue(),
  value: this. value.toValue()
}
```

smizell:
@danielgtaylor My only question is tuple vs. hash. Any specific reason to go this route?

smizell:
@pksunkara I think Daniel is away. Interested in commenting on my question above?

pksunkara:
I think hash is better.

smizell:
@pksunkara I think I agree. Reason being, if I get a tuple, I'm going to convert it to variables anyway to make it semantically understandable (i.e. I don't want `member[0]` all over my code when it could say `key`).

pksunkara:
Yup, my reasoning was the same.

smizell:
@pksunkara Would you like for me to change this, or would you like to wait for Daniel to return?

pksunkara:
I don't mind it anyway. If you are busy, I can do it later.

smizell:
@pksunkara I cannot get to it for a while today. If you end up handling it, please ping me so I don't start :) I'll do the same.

pksunkara:
I meant, I can do it next week when I finish up my current task for updating protagonist.

pksunkara:
PR opened at #76

PR #77, opened 2015-09-01
Support for custom namespaces and plugins

danielgtaylor:
This implements the interface I suggested for loading namespaces and simplifying
the minim module to just be a default registry instance. It is a breaking
change due to the renaming of public module methods for consistency with element
methods.

This continues and hopefully supersedes the discussion from #73.

cc @smizell


lib/registry.js
```diff
@@ -14,9 +13,62 @@ var BaseElement;
  */
 
 var ElementRegistry = createClass({
-  constructor: function() {
+  constructor: function(noDefault) {
     this.elementMap = {};
     this.elementDetection = [];
+    this.BaseElement = require('./primitives/base-element')(this);
```
smizell: Something I think here, it would be really nice to be able to override/extend the base element. You have set it up here where I think that's possible, though you may want to keep that in mind when designing the interface for this registry. If so, you could in a sense add some method to all elements by changing the base element.

smizell:
@danielgtaylor Nice work :) My _only_ concern with this approach is that something doesn't feel right, because it is conceptually hard to wrap your head around that Minim is itself a global registry that also provides methods for creating your own registry.

So it may not be exactly clear that when you register an element that you are registering globally for all instances of Minim. My question with that would be, should we aim to make it more obvious?

For example, with Express, you have to require Express and then start up an app, where you can "use" other modules/middleware/etc.

``` js
var express = require('express');
var app = express();
```

This feels like we are in the same situation there, but we're kind of hiding it to save a line of code. But maybe we should hide it, not sure. Just thinking out loud. What do you think?

pksunkara:
Forgot to say yesterday but lgtm.

lib/registry.js
danielgtaylor: Yeah, my goal here was twofold:
1. Make is dead simple to access the base element via `registry.BaseElement`, e.g. for subclassing.
2. Make it so you can override the base element before loading other elements.

Unfortunately, we can't just add stuff to the base element after the fact because that's not how inheritance in the js world works. We would need to modify the base element, then create all the other elements because prototypal inheritance is essentially just cloning the prototype at the time of class definition. Changes to the base prototype aren't reflected in the new class after the class has been defined. This is especially problematic for things like source maps, where every single element should be updated and I don't have a good solution for it. A namespace module may need to iterate over all currently registered elements and adjust their prototypes accordingly.

danielgtaylor:
@smizell I think you make a really good point. It is incredibly important that library authors use their own registry instances so they do not pollute the global one. I don't have strong opinions about the method that Express uses, but I have some notes:
1. Sometimes I see `var foo = require('module')();` and this pattern breaks down in ES6 modules, which becomes `import module from 'module'; const foo = module();` as its shortest approximation.
2. Express doesn't (to my knowledge) allow subclassing the `Application` class. The `express()` method acts as a constructor without exposing the class, which is something we would like to avoid since there may be legitimate use cases for subclassing a Minim element registry. We could provide both a method shortcut and expose the class.

I still like the idea of just importing and using the `minim` module as a registry, especially for app authors that don't need to worry about global registry pollution. So I guess the question becomes, is this something we solve with code (require a function call, no global registry) or documentation (tell library authors to always create a local registry). Anybody have strong opinions on this? Can we take a vote? @smizell @pksunkara

lib/registry.js
smizell: One thing that could be done would be:

``` js
var ElementRegistry = createClass({
  constructor: function(options) {
    // ...
    this.BaseElement = require('./primitives/base-element')(this, options.classMethods, options.staticMethods, options.getterSetters);
    // ...
  }

  // ....
});
```

> A namespace module may need to iterate over all currently registered elements and adjust their prototypes accordingly

I could see this as an option for people was well.

smizell:
@danielgtaylor @pksunkara As I think about it, I am starting to partially lean away from the idea of having a global registry. I think, maybe, it should be up to the user to manage a registry that is used across various modules. I wonder if doing this is too much hand holding.

---

I want to list out the spectrum of options here. 

1: One option is of course where Minim is used primarily as a global registry, though you can create your own. This is Daniel's current proposal in this PR.

``` js
var registry = require('minim');
var ElementRegistry = registry.ElementRegistry;
```

2: The other would be to still have a global registry, but not expose it as Minim module.

``` js
var registry = require('minim').globalRegistry;
var ElementRegistry = require('minim').ElementRegistry;
```

3: Last, it would be to not have a global registry at all. Something like:

``` js
var registry = require('minim').setup();
var ElementRegistry = require('minim').ElementRegistry;
```

---

For me, 2 is out because you gain very little from it. I lean toward 3 because there isn't any magic happening for the user and no surprises of some other library adding/overwriting something in the global registry. Like I said, for me, 1 mostly feels off doing it this way.

Edit: Looking back at this, I think we could have a fourth option here that is a combo of 2 and 3, but let's not go there :)

pksunkara:
I like 3 too. There will be no hidden surprises.

danielgtaylor:
@smizell @pksunkara I believe I've addressed the feedback and implemented everything from our discussion earlier. For the public record, we are going with a slight modification of number 3.

smizell:
@danielgtaylor Will look more soon, but I like what you've done with the place.

CHANGELOG.md
```diff
@@ -1,7 +1,33 @@
 # Unreleased
 
+- **BREAKING** The public interface of the `minim` module has changed significantly. List of changes:
+
+  - `ElementRegistry` has been renamed to `Namespace`.
+  - `minim` has only one public method, called `namespace`, which creates a new `Namespace` instance.
+  - `minim.convertToElement` is now `namespace.toElement`
+  - `minim.convertFromRefract` is now `namespace.fromRefract`
+  - `minim.convertFromCompactRefract` is now `namespace.fromCompactRefract`
```
smizell, line 9: These are good changes.

README.md
```diff
@@ -553,7 +555,28 @@ var elements = minim.fromCompactRefract(['category', {}, {}, [
...
+```javascript
+var minim = require('minim').namespace();
+
+// Define your plugin module (normally done in a separate file)
+var plugin = {
+  namespace: function(options) {
+    var base = options.base;
+    var ArrayElement = base.getElementClass('array');
+
+    base.register('category', ArrayElement);
```
smizell, line 574: One thought here is that there is now a lot of boilerplate that has to happen in plugins. You have to check to make sure `base` exists, and `getElementClass` in order to even register something. Would be nice to have a helper that did some of this for you.

``` js
register('category', 'array', options);
```

Of course, not necessary for this PR, just thinking out loud.

lib/namespace.js
```diff
@@ -1,22 +1,75 @@
...
- * The element registry allows you to register your own classes to be instantiated
+ * The namespace allows you to register your own classes to be instantiated
  * when a particular refract element is encountered, and allows you to specify
  * which elements get instantiated for existing Javascript objects.
  */
 
-var ElementRegistry = createClass({
-  constructor: function() {
+var Namespace = createClass({
+  constructor: function(noDefault) {
```
smizell: Are you OK with this for the future? Would it be better to have `function(options)`?

smizell:
@danielgtaylor Looks good to me, just a had a few questions. 

Also, one thing that feels off is `getElementClass`. I definitely agree with the name of this method, but wonder if we could simplify it for the user since it will be a majorly used in plugins. Not against using it at all.

README.md
danielgtaylor, line 574: I agree it would be nice. If possible though let's hold off on this for another PR. I'm honestly not sure how to make this nicer while still holding the same public interface. We might consider a helper module that namespace plugins could use or something... I think this needs more thought.

lib/namespace.js
danielgtaylor: Changing this to `options` now. Seems like we may add other stuff here in the future.

README.md
smizell, line 574: Agree. Didn't want this to be seen as a requirement, just thinking out loud.

PR #78, opened 2015-09-07
Embedded Refract

smizell:
This is mostly complete, but I want to use it as a way to talk about this format.
1. Does this make sense? Is it worth it?
2. This may need to be it's own plugin. Thoughts?

### DIFF
--- a/lib/formats/embedded-refract/from-embedded.js
+++ b/lib/formats/embedded-refract/from-embedded.js
@@ -1,0 +1,163 @@
+'use strict';
+
+var _ = require('lodash');
+
+exports.namespace = function(options) {
+  var base = options.base;
+  var parseArray;
+  var parseObject;
+  var handleNormalObject;
+  var handleRefractObject;
+  var handleContentArray;
+  var handleContentObject;
+  var overwriteFromObject;
+  var parseAndSetAsContent;
+  var getObjectKeys;
+  var isRefracted;
+  var refractProperty = '_refract';
+
+  /*
+   * Attaches parsing function to the given namespace
+   */
+  base.fromEmbeddedRefract = function(value) {
+    // Important to test for array first because they are also objects
+    // Each item of an array can be refracted
+    if (_.isArray(value)) { return parseArray(value); }
+
+    // Objects can be normal objects or refracted objects
+    if (_.isObject(value)) { return parseObject(value); }
+
+    // If we've made it here, we know the value is a simple value item with no
+    // need for extra parsing. We can treat it as a primitive value.
+    return base.toElement(value);
+  };
+
+  parseArray = function(value) {
+    var ElementClass = base.getElementClass('array');
+    var element = new ElementClass();
+
+    // Items in an array can be embedded Refract
+    _.forEach(value, function(item) {
+      element.push(base.fromEmbeddedRefract(item));
+    });
+
+    return element;
+  };
+
+  parseObject = function(value) {
+    // The refract property instructs the parser that additional parsing is
+    // needed. If this property is not set, the object is treated just like a
+    // a regular object literal.
+    if (isRefracted(value)) {
+      return handleRefractObject(value);
+    } else {
+      return handleNormalObject(value);
+    }
+  };
+
+  handleNormalObject = function(value) {
+    var ElementClass = base.getElementClass('object');
+    var element = new ElementClass();
+
+    _.forEach(getObjectKeys(value), function(key) {
+      element.set(key, base.fromEmbeddedRefract(value[key]));
+    });
+
+    return element;
+  };
+
+  handleRefractObject = function(value) {
+    var element;
+    var ElementClass = base.getElementClass(value[refractProperty].element);
+    var refractObject = value[refractProperty];
+
+    // Object elements are treated differently than other elements. If the
+    // element is not an object, just store the content (if defined).
+    if (refractObject.element === 'object') {
+      element = new ElementClass();
+      // Content as array means it is an array of member elements
+      // Otherise, parse as embedded refract and set they key equal to each
+      // value. This is essentially a shallow merge.
+      if (_.isArray(refractObject.content)) {
+        handleContentArray(element, value, refractObject.content);
+      } else {
+        handleContentObject(element, value, refractObject.content);
+      }
+    } else {
+      element = new ElementClass(refractObject.content);
+    }
+
+    parseAndSetAsContent(refractObject.attributes, element.attributes);
+    parseAndSetAsContent(refractObject.meta, element.meta);
+
+    return element;
+  };
+
+  handleContentArray = function(element, value, content) {
+    _.forEach(content, function(member) {
+      var memberKey = base.fromEmbeddedRefract(member.key);
+      element.set(memberKey.toValue(), base.fromEmbeddedRefract(member.value));
+      var memberElement = element.getMember(memberKey.toValue());
+
+      // Even member elements may be refracted, so we handle accordingly
+      if (_.has(member, refractProperty)) {
+        parseAndSetAsContent(member[refractProperty].attributes, memberElement.attributes);
+        parseAndSetAsContent(member[refractProperty].meta, memberElement.meta);
+      }
+    });
+
+    overwriteFromObject(element, value);
+
+    return element;
+  };
+
+  handleContentObject = function(element, value, content) {
+    base.fromEmbeddedRefract(content || {}).forEach(function(valueElement, keyElement) {
+      element.set(keyElement.toValue(), valueElement);
+    });
+
+    overwriteFromObject(element, value);
+
+    return element;
+  };
+
+  /*
+   * Allows for the original object's values to overwrite
+   * those of the refracted object
+   */
+  overwriteFromObject = function(element, value) {
+    // This section allows for the original object's values to overwrite
+    // those of the refracted object
+    _.forEach(getObjectKeys(value), function(key) {
+      var existingElement = element.get(key);
+
+      if (existingElement && !isRefracted(value[key])) {
+        // This allows us to retain meta and attributes
+        existingElement.content = base.fromEmbeddedRefract(value[key]).content;
+      } else {
+        element.set(key, base.fromEmbeddedRefract(value[key]));
+      }
+    });
+  };
+
+  parseAndSetAsContent = function(fromValue, toValue) {
+    if (!_.isEmpty(fromValue)) {
+      // Setting content allows us to retain meta and attributes
+      toValue.content = base.fromEmbeddedRefract(fromValue).content;
+    }
+  };
+
+  getObjectKeys = function(value) {
+    var content = _.omit(value, refractProperty);
+    return _.isEmpty(content) ? [] : _.keys(content);
+  };
+
+  isRefracted = function(value) {
+    if (_.isObject(value) && _.has(value, refractProperty)) {
+      return true;
+    }
+    return false;
+  };
+
+  return base;
+};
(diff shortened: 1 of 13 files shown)


pksunkara:
@smizell This is awesome. The tests are looking really solid. But I see that they are not completely passing though.

I still have to go through the actual code `from-embedded.js` and `to-embedded.js` and will do that later.

lib/formats/embedded-refract/from-embedded.js
```diff
@@ -0,0 +1,163 @@
...
+    var content = _.omit(value, refractProperty);
+    return _.isEmpty(content) ? [] : _.keys(content);
+  };
+
+  isRefracted = function(value) {
+    if (_.isObject(value) && _.has(value, refractProperty)) {
+      return true;
+    }
+    return false;
+  }
```
pksunkara: Lint is complaining about missing semicolon here.

smizell:
@pksunkara Thanks :) Getting a successful build now.

lib/formats/embedded-refract/from-embedded.js
```diff
@@ -0,0 +1,163 @@
...
+    var refractObject = value[refractProperty];
+
+    // Object elements are treated differently than other elements. If the
+    // element is not an object, just store the content (if defined).
+    if (refractObject.element === 'object') {
+      element = new ElementClass();
+      // Content as array means it is an array of member elements
+      // Otherise, parse as embedded refract and set they key equal to each
+      // value. This is essentially a shallow merge.
+      if (_.isArray(refractObject.content)) {
```
pksunkara, line 81: Can't the `content` here be refracted for `array` element which will lead to problems?

pksunkara, line 81: Nevermind. Looks like it is covered.

pksunkara:
Review done. Everything looking good.

smizell:
@pksunkara Do you think this should be part of the main Minim app? I'm thinking it should be its own library, and that we pull it and the compact refract parsers/serializers out as well. 

Does seeing this format in action make sense? Do you think it's helpful? It's actually one of the slimmest of all our formats I think as you only include the bare minimum into the representation. Plus, we have one format that resembles XML (compact), one that is DOM-like (full version), and now one that is JSON-like in this embedded version.

pksunkara:
Since we are supporting compact format in minim, I think this should be in minim too.

danielgtaylor:
@smizell Embedded refract looks neat and useful. Some thoughts:
- I don't mind if we split the compact and embedded formats into their own packages to be loaded with `namespace.use(...)`.
- All existing elements should be updated, not just the base element / array / object. See this bug:
  
  ``` js
  var minim = require('minim').namespace();
  var NumberElement = minim.getElementClass('number');
  var test = (new NumberElement()).fromEmbeddedRefract(1); // <=== CRASH!
  ```
  
  Javascript inheritance is not like other languages. Modifying the base type after a prototype using the base has already been created will **not** update the already created type. Prototypal inheritance is like copying the prototype over at class definition time.
  
  This is especially important when the embedded refract serializer is loaded after other namespaces may have been loaded. Ideally you want to loop through all existing elements and attach the `to`/`from` methods to them based on (I guess) whether they inherit from `ObjectElement`, `ArrayElement`, or `BaseElement`.
- I am **concerned** about elements defined in loaded namespaces that inherit from `BaseElement`. If they have custom serialization/loading behavior then the element author now needs to add yet another two methods to handle embedded refract. It's the same issue for compact refract and I wish I had a good answer to the problem, but we can't easily convert between serialization formats automatically. I've spent longer than I want to admit trying to figure out a good solution.
- Can we auto-detect the format? I.e, as a tooling author can I just load in whatever format I get from user input and stuff just works? Do I need to worry about detection? Seems like it might be hard with this embedded refract but it's something that would be really nice to have.

smizell:
@danielgtaylor Thanks for all of this :)

Regarding this example:

``` js
var minim = require('minim').namespace();
var NumberElement = minim.getElementClass('number');
var test = (new NumberElement()).fromEmbeddedRefract(1); // <=== CRASH!
```

Shouldn't users be going through the namespace to do this? 

``` js
var test = minim.fromEmbeddedRefract(1);
```

I actually think we should move any of these kinds of `from` methods away from the elements. This would mean you don't have any `fromEmbeddedRefract` on any element, even object and arrays. Thoughts?

---

> I am concerned about elements defined in loaded namespaces that inherit from BaseElement. If they have custom serialization/loading behavior then the element author now needs to add yet another two methods to handle embedded refract. It's the same issue for compact refract and I wish I had a good answer to the problem, but we can't easily convert between serialization formats automatically. I've spent longer than I want to admit trying to figure out a good solution.

Can you elaborate a little more here? I'm not sure of the issue. Is it that we are adding methods to all of the elements? Or that there is no standard interface for converting between formats?

If the latter, we could have some adapter registry like in Fury. This was why I was hoping we could come up with some pattern for this kind of registry for serializers so we could use it in Minim for this very thing. 

If I missed your reason altogether, sorry about that :)

---

> Can we auto-detect the format?

Unfortunately not if we include embedded Refract in the detection code. If it is, you would have to rely on the media type. But if it's not, you can detect whether it's normal or compact Refract (one is always an object, the other is always an array).

With that said, Embedded Refract is actually the only format that has a rule for detecting if it actually is embedded refract. If it valid JSON, it's valid embedded refract, and if it has `_refract`, then it is using the special refract object. With all others, there is no way to distinguish between plain JSON and Refract.

danielgtaylor:
@smizell 

> Shouldn't users be going through the namespace to do this?

Maybe.

---

> I actually think we should move any of these kinds of `from` methods away from the elements. This would mean you don't have any `fromEmbeddedRefract` on any element, even object and arrays. Thoughts?
> ...
> Can you elaborate a little more here? I'm not sure of the issue. Is it that we are adding methods to all of the elements? Or that there is no standard interface for converting between formats?

I'll try. The Refract specification defines a base element, and a set of elements built on top of the base that handle basic types. Refract also supports namespaces, which are in essence a logical grouping of elements that (directly or indirectly) inherit from the base element with custom behavior. The custom behavior of such elements can include how they are parsed and serialized. Their only coupling to the Refract base spec is through either the base element directly, or indirectly through an explicit dependency on another element that must exist for the namespace to work.

In order to keep the above separation of concerns it necessarily requires that namespaces define how to parse and serialize the elements they contain for each serialization format they support. This could be through `fromRefract` methods on elements or through some other mechanism that links the element name to a serialization function. The alternative is to couple the shared parser/serializer with the elements (and thus namespaces) that it explicitly supports, which to me seems like a bad design decision. Any namespace update or new namespace would require adding support in the shared serializer, for example.

With all that said, the problem:
1. A new serialization format requires updating all elements in all current namespaces to support the format before it can be used.
2. A new element must be made to support all current serialization formats before it can be used.

> If the latter, we could have some adapter registry like in Fury. This was why I was hoping we could come up with some pattern for this kind of registry for serializers so we could use it in Minim for this very thing.

The problem is that the serialization is domain-specific knowledge coupled to the namespace, because an element's `content` can by anything. I don't think it's possible to do what you are suggesting, but would love to be proven wrong. Am I wrong about these statements below?
1. Each element must know how to parse and serialize its `content`
2. Each element must know how to parse and serialize its `content` from/into each serialization format
3. The `content` and `attributes` of elements may contain a mixture of refract elements and non-refract data structures like plain js arrays / objects.
4. You cannot reliably guess when something inside of `content` is a refract element.
5. You cannot reliably traverse the `content` unless you have the domain-specific knowledge about the element.

Am I missing something here?

---

The other issue is the one of inheritance in javascript. Anytime you want to update all elements, you can't just update the base element. You must loop through all registered elements and update each one's prototype. This is just an implementation issue and isn't a major problem for me because it is trivial to fix.

smizell:
In the beginning, when we were starting with Refract, we actually did not allow this mixed situation with attributes. This meant that `attributes` was either a normal object or an array of member elements and that's it. This got some pushback and we decided that it was up to the user to rely on the namespace definition for what values were or were not refracted. Hence, as you said, we now have to mostly rely on domain knowledge for the current serialization format.

But herein lies the difficulty of the discussion, as we really have to narrow the discussion to being about the normal Refract serialization format. In **that** format, when parsing it, you must rely on the definition. In embedded Refract, this is not necessarily the case though, as a parser can distinguish between what is or is not refracted based on the presence of the `_refract` property.

Edit: this is true for the compact version as well.

---

> A new serialization format requires updating all elements in all current namespaces to support the format before it can be used.

I'm wondering if this is something we need to move away from. I'm wondering if we need something like:

``` js
var number = minim.fromFormat('embedded', 1);
var serialized = minim.toFormat('compact', number);
```

This means you would register these adapters (serializers/deserializers) and use the main library for converting to a from formats, which is similar to Fury.js. If this is the case, you would not have any of this logic on the actual elements. These adapters would simply know how to convert primitive elements, and also allow for registering in _those_ adapters any domain knowledge where applicable. This seems like it could be a better separation than putting all of this logic in the element itself.

---

Now let me go through these comments and add some thoughts. I think these comments you've made are specific to the normal and compact formats themselves and not to the internal class structure of Minim. To me, when thinking about this, it's important to separate Minim's internal state from the representation of the internal state (the formats).

> Each element must know how to parse and serialize its content

Not necessarily. These adapters must know how to parse all primitive elements, and then any additional parsing will need to be handled in a domain-specific way. In that case, that information may go on the element, but wondering if we can (a) register it somehow on the adapter or (b) provide a format-agnostic way of expressing that domain knowledge. That, of course, would be the namespace definition of the the element, which we could use :)

> Each element must know how to parse and serialize its content from/into each serialization format

As I mentioned above, not necessarily. I think we can move some of this functionality away from the actual elements.

> The content and attributes of elements may contain a mixture of refract elements and non-refract data structures like plain js arrays / objects.

This is in the serialization format, yes. Once inside Minim, though, everything is an element and there is no mixture. Attributes themselves are Object Elements, and any value in that is an instance of an element. 

> You cannot reliably guess when something inside of content is a refract element.

The embedded version is the only version that allows for this, as you know it's been refracted if there is the presence of the reserved keyword. In other versions, there is no reserved keyword, which means that anything is free game. You could try to guess if something is refracted, but you would never really know.

> You cannot reliably traverse the content unless you have the domain-specific knowledge about the element.

I think this too is related to the formats other than the embedded version. But in talking about those, you are correct.

---

> The other issue is the one of inheritance in javascript.

I agree here. I think as time goes on, we'll want a function to extend all elements in the namespace. We can use a similar API that we have.

``` js
namespace.extendElements(instanceMethods, staticMethods, gettersAndSetters);
```

smizell:
@pksunkara @danielgtaylor I've added some info to the README. If we think this is good for now, we can merge this in and start tinkering with it.

Regarding this from @danielgtaylor:

``` js
var minim = require('minim').namespace();
var NumberElement = minim.getElementClass('number');
var test = (new NumberElement()).fromEmbeddedRefract(1); // <=== CRASH!
```

I think I'd like to hold off on adding this to these other elements. Right now, users should use `minim.fromEmbeddedRefract(value)`, and also per comments above, I think we can pull everything from the actual elements.

danielgtaylor:
@smizell read through your comments and I mostly agree. This PR looks okay to merge from my end, just make sure to update the changelog as well.

I'll keep thinking about and playing with some ideas and see if I can come up with any proposals for making this better / easier.

smizell:
@danielgtaylor Changelog added, commits squashed.

pksunkara:
Read through discussion and reviewed the code. I don't see anything blocking here.

PR #80, opened 2015-09-09
Add namespace.elements

danielgtaylor:
As suggested on another [code review](https://github.com/refractproject/minim-parse-result/pull/1#discussion_r38936177):

Add an `elements` property to the `Namespace` class which returns an object of PascalCased element name keys to registered element class values. This allows for ES6 use cases like:

``` js
const {StringElement, ArrayElement, ObjectElement} = namespace.elements;
```

Possible future improvements:
- Pascal casing mechanism may need to be updated if we ever use non-camel-cased element names.
- Cache the value until an element is registered or unregistered?
  - Build the cache as elements are added or on demand when `namespace.elements` is used?

cc @smizell

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# Unreleased
+
+- Add an `elements` property to the `Namespace` class which returns an object of PascalCased element name keys to registered element class values. This allows for ES6 use cases like:
+
+  ```js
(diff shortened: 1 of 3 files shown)


pksunkara:
> Build the cache as elements are added or on demand when namespace.elements is used?

I don't think we need eager loading here

PR #85, opened 2015-10-26
Allow for load when using a plugin

smizell:
There are many times where a Minim library may not be implementing a namespace.
When this is the case, the `namespace` property does not make semantic sense
for the plugin. This commit implements a `load` property that works just like
the `namespace` property works.

### DIFF
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -31,7 +31,7 @@ var Namespace = createClass({
     if (plugin.namespace) {
       plugin.namespace({base: this});
     }
-    else if (plugin.load) {
+    if (plugin.load) {
       plugin.load({base: this});
     }
     return this;
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -25,10 +25,15 @@ var Namespace = createClass({
   },
 
   /*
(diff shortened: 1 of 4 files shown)


lib/namespace.js
```diff
@@ -25,10 +25,15 @@ var Namespace = createClass({
...
   /*
-   * Use a namespace plugin.
+   * Use a namespace plugin or load a generic plugin.
    */
   use: function(plugin) {
-    plugin.namespace({base: this});
+    if (plugin.namespace) {
+      plugin.namespace({base: this});
+    }
+    else if (plugin.load) {
```
kylef: What about a plugin that both provides a namespace and also wants to do something on load?

Perhaps make this two if statements instead of an `else if`.

smizell: I think this makes sense. I think if someone used both but only saw `namespace` used, it would be confusing.

smizell:
@kylef Comment addressed.

danielgtaylor:
LGTM :+1:

pksunkara:
:+1:

PR #89, opened 2015-11-25
Implement support for optionally refracted metadata

danielgtaylor:
This is a stab at supporting optional refracting like you see with source maps in Minim, particularly in the `meta` property of an element. This means we will now support `element.meta.id`, `element.meta.title` and `element.meta.description` metadata that includes source maps, both for serialization and for loading from refract or compact refract.

There is a (tiny) chance of false positives with the heuristics during loading. It is unlikely since we control the structure of what goes into `meta`.

If this PR is merged, it unblocks basic source map support in Minim element metadata, which is currently just being thrown away during serialization and causes fairly useless behavior when loading.

cc @smizell

### DIFF
--- a/lib/primitives/base-element.js
+++ b/lib/primitives/base-element.js
@@ -42,9 +42,10 @@ module.exports = function(registry) {
 
     toRefract: function(options) {
       var attributes = this.convertAttributesToRefract('toRefract');
+      var meta = this.convertMetaToRefract('toRefract');
       var initial = {
         element: this.element,
-        meta: this.meta.toValue(),
+        meta: meta,
         attributes: attributes,
         content: this.content
       };
@@ -53,7 +54,27 @@ module.exports = function(registry) {
 
     toCompactRefract: function() {
       var attributes = this.convertAttributesToRefract('toCompactRefract');
-      return [this.element, this.meta.toValue(), attributes, this.content];
+      var meta = this.convertMetaToRefract('toCompactRefract');
+      return [this.element, meta, attributes, this.content];
+    },
+
+    /*
+     * Converts everything in `meta` into values, unless any top-level element
+     * contains additional metadata or attributes, in which case it gets
+     * refracted.
+     */
+    convertMetaToRefract: function(functionName) {
+      var meta = {};
+      this.meta.keys().forEach(function (key) {
+        var item = this.meta.get(key);
+        if (item.meta.keys().length || item.attributes.keys().length) {
+          // Additional information, so we need to refract this element!
+          meta[key] = item[functionName]();
+        } else {
+          meta[key] = item.toValue();
+        }
+      }, this);
+      return meta;
     },
 
     /*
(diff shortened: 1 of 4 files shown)


lib/primitives/base-element.js
```diff
@@ -116,10 +137,22 @@ module.exports = function(registry) {
...
     fromRefract: function(doc) {
-      this.meta = doc.meta;
       this.attributes = doc.attributes;
       this.content = doc.content;
 
+      if (doc.meta) {
+        Object.keys(doc.meta).forEach(function (key) {
+          var value = doc.meta[key];
+
+          // TODO: Warning - this may raise a false positive :-(
```
smizell, line 147: @danielgtaylor Is this saying:

If there is an `element` property and either `meta` or `attributes`, consider it as a Refract object?

danielgtaylor, line 147: Yep. I'm open to suggestions for how to make the heuristic better.

smizell, line 147: Let's go with this, add note to the README of Minim for this special handling, and consider at a later date moving it to Refract serialization instructions.

smizell, line 147: Question: do you want to include a check for `content` since you are doing both `meta` and `attributes`?

lib/primitives/base-element.js
```diff
@@ -132,10 +165,22 @@ module.exports = function(registry) {
...
-      this.meta = tuple[1];
       this.attributes = tuple[2];
       this.content = tuple[3];
 
+      if (tuple[1]) {
+        Object.keys(tuple[1]).forEach(function (key) {
+          var value = tuple[1][key];
+
+          // TODO: Warning - this may raise a false positive :-(
+          if (value && value.length === 4 && _.isString(value[0]) && _.isObject(value[1]) && _.isObject(value[2])) {
```
smizell, line 176: This one concerns me more than the full serialization because it will create a broader signature for Refract. This will allow an array of four objects to be considered to be Compact Refract. I wonder if we could leave this one out for the time being.

lib/primitives/base-element.js
danielgtaylor, line 147: After our chat today I've updated this to only check `element`, since that is all that is required according to the spec. Also updated the README with info.

lib/primitives/base-element.js
danielgtaylor, line 176: I'm torn on this one. It might be nice to include it so that the behavior is the same for most cases, otherwise you can't load refract with sourcemaps if you are coming from compact refract.

smizell:
@danielgtaylor What about the `attributes` of an element? Does this allow for finding Refracted values in the `attributes`?

smizell:
@danielgtaylor I've accepted this, but had a question about whether this works for attributes as well. If not, then I don't think it will work for source maps, will it? If not, please disregard this questions. :)

smizell:
I'm guessing you will still need to use the special helpers for designating things in attributes, so it seems. We could use your code here in the future to get away from that.

PR #91, opened 2015-12-01
Add Hyperlinking

smizell:
Adds hyperlinking per RFC 0008 (see changelog).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,7 +17,8 @@
         "semi": [
             2,
             "always"
-        ]
+        ],
+        "no-use-before-define": 0
     },
     "env": {
         "node": true,
--- a/lib/primitives/base-element.js
+++ b/lib/primitives/base-element.js
@@ -141,16 +141,7 @@ module.exports = function(registry) {
       this.content = doc.content;
 
       if (doc.meta) {
-        Object.keys(doc.meta).forEach(function (key) {
-          var value = doc.meta[key];
-
-          // TODO: Warning - this may raise a false positive :-(
-          if (value.element) {
-            this.meta.set(key, registry.fromRefract(value));
-          } else {
-            this.meta.set(key, value);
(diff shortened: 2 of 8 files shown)


.eslintrc
```diff
@@ -17,7 +17,8 @@
         "semi": [
             2,
             "always"
-        ]
+        ],
+        "no-use-before-define": 0
```
smizell, line 21: I like to write functions in the order they are used rather than the order they are defined. Just a personal preference. Happy to reorder them if we should remove this.

smizell:
cc @danielgtaylor

.eslintrc
danielgtaylor, line 21: That's fine with me :+1:

lib/primitives/base-element.js
```diff
@@ -141,16 +141,7 @@ module.exports = function(registry) {
...
-          var value = doc.meta[key];
-
-          // TODO: Warning - this may raise a false positive :-(
-          if (value.element) {
-            this.meta.set(key, registry.fromRefract(value));
-          } else {
-            this.meta.set(key, value);
-          }
-        }, this);
+        traverseMeta(doc.meta, this);
```
danielgtaylor: What about the `fromCompactRefract` method below? In other cases I've been passing the deserialization function name around, e.g. `fromRefract` vs. `fromCompactRefract` in the new `traverse*` functions, then it can just be reused for both.

danielgtaylor:
Just the one comment above that I think is a blocker (due to introducing inconsistency between `fromRefract` and `fromCompactRefract`), otherwise it looks great :+1:

smizell:
You are exactly right. Will fix tomorrow.

smizell:
@danielgtaylor This looks great. Will you merge since it's mostly my code?

PR #92, opened 2015-12-02
Call toValue when adding items to the serialized array

danielgtaylor:
This is required to not break support down the line in other packages, e.g. when elements have arrays of non-refracted stuff like `element.meta.classes`.

cc @smizell


smizell:
:+1:

danielgtaylor:
Thanks! This is now released along with the `link` element support in `minim-0.13.0`!

PR #97, opened 2016-04-27
Finalize default refract serialization guideline

pksunkara:


lib/primitives/base-element.js
```diff
@@ -52,113 +59,86 @@ module.exports = function(registry) {
...
-          meta[key] = values;
-        } else {
-          meta[key] = item.toValue();
-        }
-      }, this);
-      return meta;
-    },
+    convertKeyToRefract: function (ret, key, item) {
+      if (this.shouldRefract(item)) {
+        ret[key] = item.toRefract();
```
smizell, line 64: What is `ret`? Would it make sense to spell it out?

test/subclass-test.js
```diff
@@ -87,48 +85,33 @@ describe('Minim subclasses', function() {
...
-              content: 'string2'
+              meta: {
+                name: 'Content-Type'
+              },
+              content: 'application/json'
             }
+          ],
+          sourceMap: [
+            'string1',
+            'string2'
```
smizell, line 107: Is this going to break things for our current implementations?

smizell, line 107: I guess this is the big breaking change?

smizell, line 107: Resolved in chat.

smizell:
@pksunkara Done from my end with just one small thing. Would like @danielgtaylor to do final review and merge.

lib/primitives/base-element.js
```diff
@@ -52,113 +59,86 @@ module.exports = function(registry) {
...
-      for (var i = 0; i < keys.length; i++) {
-        var key = keys[i];
-        if (this._attributeElementKeys.indexOf(key) !== -1) {
-          attributes[key] = this.attributes.get(key)[functionName]();
-        } else if (this._attributeElementArrayKeys.indexOf(key) !== -1) {
-          attributes[key] = [];
-          var items = this.attributes.get(key);
-          for (var j = 0; j < items.length; j++) {
-            attributes[key].push(items.get(j)[functionName]());
+      if (item.element === 'array') {
```
danielgtaylor, line 68: I know the original code did not do this, but should we be checking the `item.primitive()` instead of `item.element` here? Any subclass of an `array` (and later `object`) may have contents which should be refracted, right?

lib/primitives/base-element.js
```diff
@@ -52,113 +59,86 @@ module.exports = function(registry) {
...
+
+        ret[key] = values;
+        return;
       }
 
-      return attributes;
+      if (item.element === 'object') {
+        // This is an object, so we need to check if it's members contain
+        // additional information
+        // TODO
```
danielgtaylor, line 90: :+1:

danielgtaylor:
Looks great! :+1: Just had the one comment regarding `element` vs `primitive` above.

lib/primitives/base-element.js
pksunkara, line 68: That will be handled above in the `this.shouldRefract`. If any element is a superclass of `array`, it needs to be refracted since we need to print the element name.

danielgtaylor, line 68: Oh, that's a good point. Thanks for the clarification! LGTM :+1:

PR #98, opened 2016-04-28
Release 0.14.0 with fixes to previous PR

pksunkara:


lib/primitives/base-element.js
```diff
@@ -101,7 +98,7 @@ module.exports = function(registry) {
       var meta = {};
 
       this.meta.keys().forEach(function (key) {
-        this.convertKeyToRefract(meta, key, this.meta.get(key));
+        meta[key] = this.convertKeyToRefract(key, this.meta.get(key));
```
smizell, line 101: This is a good change. Nice work.

lib/primitives/base-element.js
```diff
@@ -114,13 +111,13 @@ module.exports = function(registry) {
...
       this.attributes.keys().forEach(function (key) {
-        this.convertKeyToRefract(attributes, key, this.attributes.get(key));
+        attributes[key] = this.convertKeyToRefract(key, this.attributes.get(key));
       }, this);
 
       return attributes;
     },
 
-    fromRefract: function(doc) {
+    fromRefract: function(doc, options) {
```
smizell, line 120: Naming question: is there a better name for this? This is not really for options for the functionality, but rather an object to merge into the newly-created element.

pksunkara, line 120: It is named as `options` in `toRefract` too.

smizell:
@pksunkara I just have one comment. The naming of the variable did not really convey what was going on to me, so I think it would be helpful to name it something different.

Other than that, this looks good to me. If we can address that, I'll merge.

lib/primitives/base-element.js
smizell, line 120: Ok, let's leave then.

PR #101, opened 2016-08-16
Update uptown to latest version

XVincentX:


smizell:
I've published 0.14.1 of Minim that has the updated Uptown.

But... Minim uses Lodash 3.10.1. :)

smizell:
I noticed this too late.

XVincentX:
Can we update this one as well?

smizell:
Definitely. I just didn't have time for it at the time. I'll check it tonight.

XVincentX:
Thanks a lot!

Issue #17, opened 2015-02-17
Hierarchy


smizell:
Elements should know of their children and their parents. 

Which brings me to a question I have. An `ObjectType` is basically a collection of `PropertyType` elements (which are essentially key-value pairs). To have the same format, and potentially help in future features, should `ArrayType` collections consist of some kind of `IndexType` that instead of having a key, it has an index. 

Reason being, an element needs to know where it is in the tree of elements, and without this having an explicit element (like `IndexType`), it would need to know the index for itself as well. If it did have this `IndexType`, it could know the array that it is in plus it's location without storing that.

Hope that makes sense.

@fosrias @zdne

fosrias:
Hm. Not sure that an element needs to know its parent. Why? It introduces unnecessary coupling doesn’t it?

smizell:
The benefit would be that you could pass around parts of the element tree and still know the context of that element. It would also allow for more complex lookups, like finding all elements with such-and-such attribute where the parent element is of such-and-such type.

fosrias:
I am not sure this should be accomdated but the structure, but rather every element just having a parent property that is dependency injected when instantiating the wrapping class, no?

smizell:
@fosrias Exactly. This parent relationship will not be handled by the Refract data structure, but rather by Minim. It is internal to element instance.

kylef:
This was solved by #141.

Issue #13, opened 2015-02-14
Immutability


smizell:
I am interested in making the getters and setters be immutable. This means that when you set a value, the old value is not changed, but rather a new object is returned. This requires some fancy work to pass changes up and down the hierarchy and to share the tree, which shouldn't be too hard.

When it's done, it could be a really cool library for use with React, as a component could listen in on only one part of the tree using callbacks.

fosrias:
What about performance with that overhead of duping the object and then changing?

smizell:
No need to dupe the object I don't think. Since we are not using a data
structure here and simply JS objects for each value, we can share the whole
tree and only change the one value while saving the previous.

smizell:
I'm going to table this idea for now.

smizell:
Well, I'm going to leave this open, as this may be something for future releases. I found this and it may actually be a cool fit for the internals of this if immutability is desired.

https://github.com/facebook/immutable-js

PR #106, opened 2017-04-25
Allow querying the element tree recursively

kylef:
Recursively find an element. Returns an ArrayElement containing all elements that match the given element name.

```javascript
const strings = element.findRecursive('string');
```

You can pass multiple element names to `findRecursive`. The method will search for the last given element name inside elements matching the other given element names.

```javascript
const stringInsideMembers = element.findRecursive('member', 'string');
```

**NOTE:** *The returned elements contain a `parents` property which includes all of the elements parent in the original Refract tree.*

The `parents` property is injected in just like mentioned in https://github.com/refractproject/minim/issues/17

**Recommended to review per commit.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,8 @@ element including the parents of the returned element.
 
 As an example, if I had an array element which contains a category element
 which in turn contains a string element with the content "Hello World". I can
-access the parent array and category element.
+access the parent array and category element. The parents are in closest parent
+order.
 
 ```json
 {
@@ -244,10 +245,10 @@ access the parent array and category element.
 const elements = element.findRecursive('string');
 const helloString = elements.first();
 
-// Array Element
+// Category Element
 helloString.parents[0];
 
-// Category Element
+// Array Element
 helloString.parents[1];
 ```
 
--- a/README.md
+++ b/README.md
@@ -431,6 +431,26 @@ arrayElement.forEach(function(item) {
 }); // logs each value to console
 ```
 
+##### shift
+
+The `shift` method may be used to remove an item from the start of a Minim array.
+
+```javascript
+var arrayElement = minim.toElement(['a', 'b', 'c']);
+var element = arrayElement.shift();
+console.log(element.toValue()); // a
+```
+
+##### unshift
+
+The `unshift` method may be used to inserts items to the start of a Minim array.
+
+```javascript
+var arrayElement = minim.toElement(['a', 'b', 'c']);
+arrayElement.unshift('d');
+console.log(arrayElement.toValue()); // ['d', 'a', 'b', 'c']
+```
+
 ##### push
 
 The `push` method may be used to add items to a Minim array.
--- a/README.md
+++ b/README.md
@@ -195,13 +195,21 @@ var stringElement = minim.toElement("foobar");
 var stringElementClone = stringElement.clone();
 ```
 
-#### recursiveFind
+#### findRecursive
 
 Recursively find an element. Returns an ArrayElement containing all elements
 that match the given element name.
 
 ```javascript
-const strings = element.recursiveFind('string');
+const strings = element.findRecursive('string');
+```
+
+You can pass multiple element names to `findRecursive`. The method will
+search for the last given element name inside elements matching the other
+given element names.
+
+```javascript
+const stringInsideMembers = element.findRecursive('member', 'string');
 ```
 
 **NOTE:** *The returned elements contain a `parents` property which includes
--- a/README.md
+++ b/README.md
@@ -204,16 +204,52 @@ that match the given element name.
 const strings = element.findRecursive('string');
 ```
 
-You can pass multiple element names to `findRecursive`. The method will
-search for the last given element name inside elements matching the other
-given element names.
+You may pass multiple element names to `findRecursive`. When multiple element
+names are passed down, minim will only find an element that is found within
+the other given elements. For example, we can pass in `member` and `string` so
+that we are recursively looking for all `string` elements that are found within a
+`member` element:
 
 ```javascript
 const stringInsideMembers = element.findRecursive('member', 'string');
 ```
 
-**NOTE:** *The returned elements contain a `parents` property which includes
-all of the elements parent in the original Refract tree.*
+##### Parents
+
+Each returned element will include a new `parents` property which is an array
+element including the parents of the returned element.
+
+As an example, if I had an array element which contains a category element
+which in turn contains a string element with the content "Hello World". I can
+access the parent array and category element.
+
+```json
+{
+  "element": "array",
+  "content": [
+    {
+      "element": "category",
+      "content": [
+        {
+          "element": "string",
+          "content": "Hello World"
+        }
+      ]
+    }
+  ]
+}
+```
+
+```javascript
+const elements = element.findRecursive('string');
+const helloString = elements.first();
+
+// Array Element
+helloString.parents[0];
+
+// Category Element
+helloString.parents[1];
+```
 
 ### Minim Elements
 
--- a/README.md
+++ b/README.md
@@ -197,11 +197,15 @@ var stringElementClone = stringElement.clone();
 
 #### recursiveFind
 
-Recursively find an element.
+Recursively find an element. Returns an ArrayElement containing all elements
+that match the given element name.
 
 ```javascript
 const strings = element.recursiveFind('string');
 ```
+
+**NOTE:** *The returned elements contain a `parents` property which includes
+all of the elements parent in the original Refract tree.*
 
 ### Minim Elements
 
--- a/README.md
+++ b/README.md
@@ -195,6 +195,14 @@ var stringElement = minim.toElement("foobar");
 var stringElementClone = stringElement.clone();
 ```
 
+#### recursiveFind
+
+Recursively find an element.
+
+```javascript
+const strings = element.recursiveFind('string');
+```
+
 ### Minim Elements
 
 Minim supports the following primitive elements
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -164,7 +164,7 @@ var Namespace = createClass({
   /*
    * Convert a refract document into refract element instances.
    */
-   fromRefract: function(doc) {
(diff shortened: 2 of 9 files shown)


kylef:
⚠️ eslint wasn't happy and was running super old version which produces poor messages, I updated eslint and resolved the remaining issues in 84e141de. I discovered that newer eslint doesn't pass on Node 0.10/0.12 so I removed support for Node 0.10/0.12 in 9b5c961 ⚠️

pksunkara:
Yeah, we need to convert this library to use peasant too.

README.md
```diff
@@ -195,13 +195,21 @@ var stringElement = minim.toElement("foobar");
...
 that match the given element name.
 
 ```javascript
-const strings = element.recursiveFind('string');
+const strings = element.findRecursive('string');
+```
+
+You can pass multiple element names to `findRecursive`. The method will
+search for the last given element name inside elements matching the other
+given element names.
```
pksunkara: Can we rephrase this a bit? I don't think the given description is clear.

lib/primitives/base-element.js
```diff
@@ -35,6 +35,103 @@ module.exports = function(registry) {
...
 
+    /// Finds the given elements in the element tree
+    /// Returns ArrayElement of elements or null
+    findRecursive: function() {
+      var elementNames = [].concat.apply([], arguments);
+      var ArrayElement = registry.getElementClass('array');
+      var array = new ArrayElement();
+      var parents;
+
+      if (elementNames[elementNames.length - 1].element === 'array') {
```
pksunkara, line 47: Since all `elementNames` are strings, wouldn't trying to do, `"random string".element` give an error?

lib/namespace.js
```diff
@@ -164,7 +164,7 @@ var Namespace = createClass({
   /*
    * Convert a refract document into refract element instances.
    */
-   fromRefract: function(doc) {
+  fromRefract: function(doc) {
```
kylef, line 167: the eslint update picked up on this incorrect spacing.

lib/primitives/base-element.js
```diff
@@ -35,6 +35,104 @@ module.exports = function(registry) {
...
+    /// Returns ArrayElement of elements or null
+    /// findRecursive: function(names..., parents)
+    findRecursive: function() {
+      var elementNames = [].concat.apply([], arguments);
+      var ArrayElement = registry.getElementClass('array');
+      var array = new ArrayElement();
+      var parents;
+
+      if (elementNames[elementNames.length - 1].element === 'array') {
+        parents = elementNames.pop().clone();
```
pksunkara, line 48: Why do we need this clone?

README.md
```diff
@@ -195,6 +195,26 @@ var stringElement = minim.toElement("foobar");
...
+
+```javascript
+const stringInsideMembers = element.findRecursive('member', 'string');
+```
+
+**NOTE:** *The returned elements contain a `parents` property which includes
+all of the elements parent in the original Refract tree.*
```
pksunkara: We need to give an example for this `parents` property. I should have expected it but I was surprised to find that the order is from top to bottom.

pksunkara (changes requested):

lib/primitives/base-element.js
kylef: Otherwise, when we add our self (`this` element) a few lines down then we are mutating the given inherited parents array element which in turn will cause the current element to show up on other parent trees that are attach to other elements.

It was actually discussed about making getters/setters immutable too which could allow us to handle this differently in the future -> https://github.com/refractproject/minim/issues/13

lib/primitives/base-element.js
kylef: No, because `"random string".element` returns undefined and this the comparison fails. If that wasn't the case then all the tests that call `findRecursive` would fail.

README.md
kylef: fefca33

README.md
kylef: fefca33

README.md
```diff
@@ -195,6 +195,62 @@ var stringElement = minim.toElement("foobar");
...
+that we are recursively looking for all `string` elements that are found within a
+`member` element:
+
+```javascript
+const stringInsideMembers = element.findRecursive('member', 'string');
+```
+
+##### Parents
+
+Each returned element will include a new `parents` property which includes an
```
pksunkara:
```diff
- which includes an array
+ which is an array
```

pksunkara (approved):

README.md
kylef: 0978820

README.md
```diff
@@ -195,6 +195,62 @@ var stringElement = minim.toElement("foobar");
...
+```javascript
+const elements = element.findRecursive('string');
+const helloString = elements.first();
+
+// Array Element
+helloString.parents[0];
+
+// Category Element
+helloString.parents[1];
+```
```
honzajavorek: I think the enumeration would make more sense reversed. If you read it out loud: "category is the 1st parent of helloString", "array is the 2nd parent of helloString"

test/primitives/base-element-test.js
```diff
@@ -231,4 +231,158 @@ describe('BaseElement', function() {
       });
     });
   });
+
+  context('when querying', function() {
+    it('returns null when there are no matching elements', function() {
+      const element = new minim.BaseElement();
+      const result = element.findRecursive('string');
+
+      expect(result).to.equal(null);
```
honzajavorek: since this should return something like array, wouldn't it make sense to return an empty array in this case?

test/primitives/base-element-test.js
```diff
@@ -231,4 +231,158 @@ describe('BaseElement', function() {
...
+      const element = new minim.BaseElement(
+        new StringElement('Hello World')
+      );
+
+      const result = element.findRecursive('string');
+
+      expect(result.element).to.equal('array');
+      expect(result.toValue()).to.deep.equal(['Hello World']);
+    });
+
```
honzajavorek, line 254: what would be a result of `new StringElement('Hello World').findRecursive('string')`?

test/primitives/base-element-test.js
```diff
@@ -231,4 +231,158 @@ describe('BaseElement', function() {
...
+      const result = element.findRecursive('string');
+
+      expect(result.element).to.equal('array');
+      expect(result.toValue()).to.deep.equal(['Hello World']);
+
+      const helloElement = result.first();
+      const parentIDs = helloElement.parents.map(function (item) {
+        return item.id.toValue();
+      });
+      expect(parentIDs).to.deep.equal(['Outter', 'Inner']);
```
honzajavorek: I'd prefer

```diff
- ['Outter', 'Inner']
+ ['Inner', 'Outter']
```

(see my comment above)

honzajavorek:
I'm thinking out loud... What would be the best way to implement [something like this](https://github.com/apiaryio/dredd-transactions/blob/master/src/detect-transaction-examples.coffee#L41:L42)?

That means - is it possible to have a transition as an input, then lookup all requests and responses it contains and sort them according to their position in the original document? I guess with `findRecursive` it would be fairly simple... (pseudocode, I'm not very familiar with minim):

```js
requests = transitionElement.findRecursive('request')
responses = transitionElement.findRecursive('response')

elements = requests.concat(responses)
elements = elements.map(function (element) {
  return {
    type: element.name,
    transaction: element.parents.findRecursive('transaction')[0],
    position: getPositionInOriginalDocument(element)
  }
}).sort(function (x, y) {
  return x.position - y.position
})

function getPositionInOriginalDocument(element) {
  sourceMaps = element.findRecursive('sourceMap')  // returns source maps of any elements within the given request or response element
  positions = sourceMaps.map(function(sourceMap) { return toPositionNumber(sourceMap) })
  return Math.min.apply(null, positions)
}
```

It's possible to lookup sourceMaps like this, even though they're in attributes, right?

pksunkara:
This `findRecursive` is not a hammer. You can do `element.sourceMap`. And you can do `findByElement('transaction')` since you don't need it to be recursive and don't need parents.

honzajavorek:
> You can do `element.sourceMap`.

I remember I didn't want `element.sourceMap` for some reason. It was missing or misleading in some cases, so I took all source maps of child elements of request or response and found the one with the lowest position in the document. That worked. That's why I needed to get all source maps within a certain element and that's why refractQuery originally didn't solve my use case (it wasn't able to traverse attributes).

> you can do `findByElement('transaction')`

How come `request.findByElement('transaction')` would work? What I'm trying to achieve on that line is: "I have a request element. Give me the closest parent transaction element." I thought `findByElement()` searches through children.

pksunkara:
No, `element.parents.findByElement('transaction')`. And it's not that sourceMaps are missing sometimes, it's that they don't exist for some things. It's deterministic, so I think code should be too instead of searching for nearest source map.

honzajavorek:
@pksunkara If both `+ Request` and `+ Response` sections have their direct source maps, I'm okay. But something tells me it wasn't like that in the past. That's why I have [these tests](https://github.com/apiaryio/dredd-transactions/blob/master/test/unit/detect-transaction-examples-test.coffee).

If they do not provide direct source map attribute, the combinations of what they can contain are endless and in reality, I do not really care what exactly they contain - I need just the earliest thing in the document to detect where the request or response starts, to determine their position. But if there's something like `traverseAllChildren`, I'm okay to implement the recursive source maps lookup by myself.

pksunkara:
>  the combinations of what they can contain are endless

They are not endless. The request/response have the source maps directly. I think the tests you have are a relic of APIB AST.

honzajavorek:
They're not, they're specifically for Refract. But if both req and res have
their own source maps, then I'm fine.

kylef:
@honzajavorek @pksunkara Thanks for the feedback, I've flipped order of parents so they are closest first in a6d837f and I've made it so that empty array is returned instead of null in 6c5b552.

I've also added a shift and unshift methods for ArrayElement in c18eaa4 because I needed to use this behaviour during the other two commits.

@pksunkara can you please review changes and merge if you are happy.

Issue #59, opened 2015-06-28
Add remove method


smizell:
The remove method would be for array and object elements. Remove would not destroy the instance, but simply remove it from the content of the item.

PR #107, opened 2017-04-28
feat: Added remove

pksunkara:


lib/primitives/object-element.js
```diff
@@ -45,6 +45,14 @@ module.exports = function(BaseElement, ArrayElement, MemberElement) {
       );
     },
 
+    delete: function (name) {
+      this.content = this.content.filter(function (item) {
+        return item.key.toValue() !== name;
+      });
+
+      return this;
```
kylef: Why would it return `this`? Why not return the deleted element like the other collection removal APIs in JS.

pksunkara: Chaining. Check the tests. And I wanted to talk to you about this today.

lib/primitives/object-element.js
```diff
@@ -45,6 +45,14 @@ module.exports = function(BaseElement, ArrayElement, MemberElement) {
       );
     },
 
+    delete: function (name) {
```
kylef: `name`? Shouldn't this be key of the element?

README.md
```diff
@@ -523,6 +523,18 @@ var objectElement = minim.toElement({ foo: 'bar' });
 var keys = objectElement.keys() // ['foo']
 ```
 
+##### delete
```
kylef: I think `remove` is a better name for this method as we're removing an element from the object. Deleting would imply the element is gone, but if I had a prior reference to it I can keep using the element. It is not "deleted".

README.md
```diff
@@ -523,6 +523,18 @@ var objectElement = minim.toElement({ foo: 'bar' });
...
+
+The `delete` method deletes a key from the `ObjectElement` instance.
```
kylef: I think it's worth mentioning it's the key's content (value) and not the key (element) itself. This method does not accept the key only the keys value.

kylef (commented):
Do you want to make this this method apply to arrays to so we can close off https://github.com/refractproject/minim/issues/59 too?

README.md
pksunkara: What? I don't understand the above sentence

kylef: Both the following are not supported:

```js
object.delete(memberElement);
object.delete(keyElement);
```

Only this works:

```js
object.delete(keyValue);
```

pksunkara: Ah. Shouldn't the example give that away? All the other examples do the same thing. Let's not confuse the users by introducing an explanation on not supporting elements.

lib/primitives/object-element.js
pksunkara: It's the name of the key. This is the function signature being used everywhere. I didn't want to change it.

pksunkara:
@kylef Updated.

lib/primitives/object-element.js
kylef: I don't think it makes sense to return the collection. Unless it was returning a new collection with the removed item (non mutating).

```javascript
const newObject = object.removing('somekey');
```

I think its far more likely you would want to chain the deleted element itself. This remove method doesn't offer the caller any feedback on whether the element was removed from the collection.

```javascript
if (object.remove('request')) {
    console.log('there was a request and we removed it');
}
```

```javascript
const response = object.remove('response');
if (response) {
    console.log('there was a response and we removed it');
    // do something with response.statusCode
}
```

kylef: Can you please add a test case for removing an element that doesn't exist too.

kylef:
@pksunkara can you rebase please.

pksunkara:
Done.

kylef (approved):

PR #113, opened 2017-05-23
Support `parent` property on all elements

kylef:
This PR adds a `parent` property on an element which is set to a parent when a child is added to a collection. For example, when I set a key of a member the keys parent is set to the member.

See README and CHANGELOG changes in PR diff for more details.

Implements #17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+# Master
+
+## Breaking
+
+- `findRecursive` now returns an array instead of an `ArrayElement`.
+- The parents on the elements returned by `findRecursive` are no longer an
+  `ArrayElement`.
+
+## Enhancements
+
+- Element now contains a `parent` property which may be set to an elements
+  parent element.
+- Element has a `parents` property returning all of the parent elements of an
+  element.
+
 # 0.16.0 - 2017-05-04
 
 ## Breaking
--- a/lib/primitives/base-element.js
+++ b/lib/primitives/base-element.js
@@ -37,47 +37,29 @@ module.exports = function(registry) {
 
     /// Finds the given elements in the element tree
     /// Returns ArrayElement of elements
-    /// findRecursive: function(names..., parents)
+    /// findRecursive: function(names...)
     findRecursive: function() {
       var elementNames = [].concat.apply([], arguments);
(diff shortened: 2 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+# Master
+
+## Breaking
+
+- `findRecursive` now returns an array instead of an `ArrayElement`.
```
kylef, line 5: This change is necessary because otherwise adding the children to a new array element will change the parent and have undesired effects. I don't see much motivation of find recursive (and parents below) methods for returning `ArrayElement` instead of an array. There is no meta or attributes set, and other high-order functions such as `map` and `reduce` deal in arrays and not ArrayElement.

kylef: Also to note, `children` on array element returns array too. So IMO we should return array here for consistency.

kylef:
This does leave one problem, that if I add an element to a collection, and then re-add it to another collection. The elements parent is only for the last collection. I was thinking about this being a known limitation but not blocking this PR, something to :thinking: in the future.

lib/primitives/base-element.js
```diff
@@ -37,48 +37,31 @@ module.exports = function(registry) {
 
     /// Finds the given elements in the element tree
     /// Returns ArrayElement of elements
-    /// findRecursive: function(names..., parents)
+    /// findRecursive: function(names...)
     findRecursive: function() {
       var elementNames = [].concat.apply([], arguments);
       var ArrayElement = registry.getElementClass('array');
```
pksunkara: Don't need this.

pksunkara:
What about the parent of a member element itself? Shouldn't it be set to object?

kylef:
> What about the parent of a member element itself? Shouldn't it be set to object?

It should be, if you add member to an object. Since object inherits from array, array will configure the child.

CHANGELOG.md
kylef: Actually, I realise this may be problematic as you will loose access to the convenience element searching methods.

pksunkara: True, We need it to be element array.

pksunkara:
> It should be, if you add member to an object. Since object inherits from array, array will configure the child.

Can you add tests for this?

PR #114, opened 2017-05-28
Untangling spaghetti

kylef:
This pull request hugely refactors Minims architecture to decouple some of the components. Elements are no longer responsible for serialisation and now don't even know what a namespace is. Serialisation is now [composable](https://en.wikipedia.org/wiki/Composability).

**I would recommend reviewing per-commit.**

- Elements no longer know about namespaces.
- There is a new `KeyValuePair` type which removes some special-cases of the MemberElement which used an object as its content. We no longer use an arbitrary object with stringy keys.
- bd63f1e1c5fdb33cf341a7da784e382bb3646d47  - `Element.toValue()` is now generic and can handle all valid elements. This means that we do not have to customise `toValue` in subclasses. We can expect `toValue` to "just work" for any element type. We do not need upfront knowledge of element semantics to make `toValue` work.
- *JSON Serialisation is now decoupled* from the Elements themselves and is a separate construct
    - Namespaces provide the existing convince functionality of `toRefract`/`fromRefract` along with a `serialiser` property to easily create a JSON serialiser.
    - It is now easier to extend and customise the serialiser. A subclass of the `JSONSerialiser` can be made overriding functionality. This make it easier for us to provide both the new serialisation format and the existing serialisation format at the same time. We can provide an option of the `JSONSerialiser` to change formats.
      ```javascript
      const serialiser = new JSONSerialiser(namespace, {/* options to optionally refract elements */});
       ```
- Clone no longer relies on knowledge of namespace and JSON serialisation now does direct object copies instead of serialising and deserialising. This should also give us a performance improvement.

**I would recommend reviewing per-commit.**

This PR solves problems outlined in https://github.com/apiaryio/dredd-transactions/pull/77#discussion_r118586772.


test/primitives/array-element-test.js
```diff
@@ -48,12 +48,6 @@ describe('ArrayElement', function() {
       });
     });
 
-    describe('#toValue', function() {
-      it('returns the array', function() {
-        expect(arrayElement.toValue()).to.deep.equal(['a', true, null, 1]);
-      });
-    });
-
```
pksunkara, line 56: Please don't remove this test since the other tests you have written doesn't use `ArrayElement` to instantiate.

pksunkara: Similarly, all the other tests.

test/serialisers/json.js
```diff
@@ -0,0 +1,273 @@
...
+      expect(object).to.deep.equal({
+        element: 'string',
+        meta: {},
+        attributes: {
+          thread: 123
+        },
+        content: 'Hello World'
+      });
+    });
+  });
```
pksunkara, line 174: Object serialisation?

test/serialisers/json.js
```diff
@@ -0,0 +1,273 @@
...
+            content: 'hello'
+          }
+        }
+      });
+
+      const attribute = element.attributes.get('thing');
+      expect(attribute).to.be.instanceof(minim.elements.String);
+      expect(attribute.content).to.equal('hello');
+    });
+
```
pksunkara, line 326: Object Deserialise?

lib/primitives/base-element.js
```diff
@@ -29,7 +29,33 @@ module.exports = function(registry) {
...
+      var copy = new this.constructor();
+
+      copy.element = this.element;
+
+      if (this.meta.length) {
+        copy._meta = this.meta.clone();
+      }
+
+      if (this.attributes.length) {
+        copy._attributes = this.attributes.clone();
```
pksunkara: Are you sure you want to set `_meta` and `_attributes` like this?

pksunkara:
I haven't reviewed the last commit yet. Here are my comments until then. Also, I have to ask this. What about embedded-refract format?

lib/primitives/base-element.js
kylef: Yes, since [meta/attributes setter](https://github.com/refractproject/minim/blob/02d426f8585810e79a3e8329bbeccc3a186fabdf/lib/primitives/base-element.js#L293) on BaseElement doesn't replace the ObjectElement but instead passes the given value to `objectElement.set()`. In turn, [`set`](https://github.com/refractproject/minim/blob/02d426f8585810e79a3e8329bbeccc3a186fabdf/lib/primitives/object-element.js#L74) doesn't expect to receive an `ObjectElement` in this way and will fail. You can run the test suite with this line changed to `copy.attributes = this.attributes.clone();` if you want to dig into this further.

test/serialisers/json.js
kylef: 8d0be1519c1e5940af92c4dfa33489565d9f076d

test/serialisers/json.js
kylef: 8d0be1519c1e5940af92c4dfa33489565d9f076d

kylef:
> What about embedded-refract format?

I did not touch embedded refract in this PR, since the way it is injected into BaseElement it wasn't required to touch it during this refactoring. However I do agree it does need changing in the future.

I have never seen embedded Refract used anywhere and the format, perhaps we can remove this format in favour of the JSON serialiser?

kylef:
I've added commit ef416b4 which makes the test coverage 100% by testing some previously untested functionality.

#### `ef416b4`

screen shot 2017-05-29 at 22 19 48

#### Master

screen shot 2017-05-29 at 22 20 20

kylef:
Created https://github.com/refractproject/minim/pull/115 which targets this PR.

pksunkara (approved):

PR #115, opened 2017-05-29
Provide element classes in public interface

kylef:
This pull request provides the element classes as apart of the minim public interface. This makes it easier for other libraries to include and extend specific elements without having to be wrapped in a namespace.

As an example, in minim-api-description could be simplified:

##### After

```js
import { BaseElement } from 'minim';

export default class Asset extends BaseElement {}
```

```js
import Asset from './elements/asset';

export function namespace(options) {
  options.base.register('asset', Asset);
}

export default { namespace, Asset };
```

##### Before

```js
export default function (namespace) {
  class Asset extends namespace.BaseElement {}

  namespace.register('asset', Asset);
}
```

```js
import asset from './elements/asset';

export function namespace(options) {
  asset(options.base);
}

export default { namespace };
```

### DIFF
--- a/lib/minim.js
+++ b/lib/minim.js
@@ -9,3 +9,13 @@ exports.Namespace = Namespace;
 exports.namespace = function namespace(options) {
   return new Namespace(options);
 };
+
+exports.BaseElement = require('./primitives/base-element');
+exports.StringElement = require('./primitives/string-element');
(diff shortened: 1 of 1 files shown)


kylef:
Also been thinking about this, perhaps we should rename `BaseElement` to `Element`.

pksunkara (approved):

PR #120, opened 2017-06-14
Remove embedded refract

kylef:
Embedded Refract is not used anywhere, and is not specified anywhere. I think we can safely remove this Embedded Refract support given it is not used and doesn't look like it will be used.


pksunkara:
It is actually specced out in the main refract spec repo, but yes, we can remove it from minim since it's not being used.

pksunkara (approved):

PR #122, opened 2017-06-23
Convert JS objects during "refraction" of JavaScript types

kylef:
The `refract` function is supposed to take any JS type and convert it to a Refract element. The case where the given value was a JS Object it wasn't correctly handled.

### DIFF
--- a/lib/refraction.js
+++ b/lib/refraction.js
@@ -2,13 +2,17 @@
 
 /// Refracts a JSON type to minim elements
 function refract(value) {
+  var Element = require('./primitives/element');
   var ArrayElement = require('./primitives/array-element');
   var NumberElement = require('./primitives/number-element');
   var StringElement = require('./primitives/string-element');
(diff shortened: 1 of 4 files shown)


lib/refraction.js
```diff
@@ -18,6 +22,14 @@ function refract(value) {
...
   } else if (Array.isArray(value)) {
     return new ArrayElement(value.map(refract));
+  } else if (typeof value === 'object') {
+    var element = new ObjectElement();
+
+    for (var key in value) {
+      element.set(key, value[key]);
+    }
+
+    return element;
```
pksunkara, line 32: I don't see a test for this in this PR. Is this being tested elsewhere?

kylef: There wasn't any unit tests for `refract` function. I've added a full set in https://github.com/refractproject/minim/pull/122/commits/19d06d29b6f909334f0ef8860f3f8e96b37d7917.

pksunkara (approved):

PR #125, opened 2017-06-29
Add special case to serialise sourceMap as value

kylef:
Since sourceMap doesn't serialise its content as other element, it has a special rules of its own to serialise a plain array of array. Therefore we can handle the special case by calling `toValue()` on the sourceMap element when serialising sourceMaps.

This PR goes with https://github.com/refractproject/minim-parse-result/pull/20/files#r124780274.

This isn't so elegant, but I don't see a better way to solve the problem with the existing serialisation rules. This rule is already changed in the latest version of the Refract JSON Serialisation and we can remove it going forward when we update serialisation in Minim (moving current implementation to legacy serialiser).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Initialising an Element with given meta or attributes as ObjectElement is now
   supported.
 - When converting JavaScript values to Refract, objects are now supported.
+- Adds a special case to serialise sourceMap elements as values.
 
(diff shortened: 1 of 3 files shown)


pksunkara:
This is only temporary for the old version of refract, right?

pksunkara (approved):

kylef:
> This is only temporary for the old version of refract, right?

Yes, the new serialiser won't need this logic.

PR #132, opened 2017-07-19
Introduce new type to represent sub collection of elements

kylef:
This change introduces a new type (`ElementSlice`) which represents a collection of elements. This new type is NOT an element itself. An `ElementSlice` is effectively a thin wrapper around `Array` but has the capability of providing convenience filtering and querying functionality on top.

 Querying and filtering functionality of Minim will now use the `ElementSlice` type instead of `ArrayElement`. This is an important distinction against the previous implementation which returned an element when filtering. This is part of a bigger picture, for an element to keep track of its parent (https://github.com/refractproject/minim/issues/17) element and to be able to query up the element tree along with downwards.

When filtering and querying itself would return a further element, this would have the downside of the parent element changing ownership and complicates the implementation significantly when an element should keep track of its parent.

`ElementSlice` doesn't contain any specialised filtering/querying methods at the moment, I will make separate change requests for these so we can consider them one by one.

#### Element Slice

I've though long about the name for this and a slice is the best name I could come up with. I do not want to overload the term `Array` as there are already two types of array `Array`, and `minim.Array` (Element). I believe it would be further complicating to have a third concept sharing a similar name. Since `ElementSlice` is a type that represents a "Slice" of elements which are contained in a pre-existing tree, I think Slice adequately describes the concept.

This is sharing the same name for a similar concept in JavaScript [Array.prototype.slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Return_a_portion_of_an_existing_array) and a [general concept in computer science](https://en.wikipedia.org/wiki/Array_slicing).

> In computer programming, array slicing is an operation that extracts a subset of elements from an array and packages them as another array, possibly in a different dimension from the original.

### DIFF
--- a/lib/element-slice.js
+++ b/lib/element-slice.js
@@ -1,0 +1,105 @@
+'use strict';
+
+var _ = require('lodash');
+var uptown = require('uptown');
+var refract = require('./refraction').refract;
+var createClass = uptown.createClass;
+
+var ElementSlice = createClass({
+  // elements: [Element]
+  constructor: function (elements) {
+    this.elements = elements || [];
+  },
+
+  toValue: function () {
+    return this.elements.map(function (element) {
+      return element.toValue();
+    });
+  },
+
+  // High Order Functions
+
+  map: function (callback, thisArg) {
+    return this.elements.map(callback, thisArg);
+  },
+
+  // Filters the element slice
+  // Returns a new sub slice
+  filter: function (callback, thisArg) {
+    return new ElementSlice(this.elements.filter(callback, thisArg));
+  },
+
+  forEach: function (callback, thisArg) {
+    this.elements.forEach(callback, thisArg);
+  },
+
+  reduce: function (callback, initialValue) {
+    return this.elements.reduce(callback, initialValue);
+  },
+
+  includes: function(value) {
+    for (var i = 0; i < this.length; i++) {
+      var item = this.elements[i];
+      if (_.isEqual(item.toValue(), value)) {
+        return true;
+      }
+    }
+
+    return false;
+  },
+
+  // Mutation
+
+  shift: function() {
+    return this.elements.shift();
+  },
+
+  unshift: function(value) {
+    this.elements.unshift(refract(value));
+  },
+
+  push: function(value) {
+    this.elements.push(refract(value));
+    return this;
+  },
+
+  add: function(value) {
+    this.push(value);
+  },
+
+  // Accessors
+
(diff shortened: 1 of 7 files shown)


lib/primitives/element.js
```diff
@@ -94,18 +95,19 @@ var Element = createClass({
...
     var elementNames = [].concat.apply([], arguments);
-    var array = new ArrayElement();
+    var elements = new ElementSlice();
     var parents;
 
-    if (elementNames[elementNames.length - 1].element === 'array') {
-      parents = elementNames.pop().clone();
+    if (elementNames[elementNames.length - 1] instanceof ElementSlice) {
+      // clone the given parents
+      parents = new ElementSlice(elementNames.pop().map(function (e) { return e; }));
```
kylef, line 108: This looks a little messy, but it is exactly how `clone()` on element was working for an ArrayElement, by using `map` to create a copy. I have kept same behaviour for consistency.

This line and surrounding code will be removed in a future PR I am working on to add parent tracking to elements so that we don't need to track parents and make clones here.

test/element-slice-test.js
```diff
@@ -0,0 +1,145 @@
...
+
+    expect(slice.length).to.equal(1);
+    expect(shifted).to.equal(one);
+  });
+
+  it('allows unshifting an element', function () {
+    var two = new Element('two');
+    var slice = new ElementSlice([two]);
+
+    slice.unshift('one');
```
pksunkara, line 108: What if I give a refract element here directly?

lib/element-slice.js
```diff
@@ -0,0 +1,105 @@
...
+  },
+
+  // Mutation
+
+  shift: function() {
+    return this.elements.shift();
+  },
+
+  unshift: function(value) {
+    this.elements.unshift(refract(value));
```
pksunkara, line 58: I am concerned with doing `refract` here. What is value is an element already? Similarly below.

pksunkara (changes requested):

lib/element-slice.js
kylef: `refract` is effectively a noop when the element is already Refracted (https://github.com/refractproject/minim/blob/master/lib/refraction.js#L14), it will just return the element.

pksunkara (approved):

pksunkara:
Huh. The rebase option was on. So, the PR got rebased and merged.

PR #137, opened 2017-07-31
Implement parent element tracking

kylef:
Supersedes https://github.com/refractproject/minim/pull/58

This PR adds `parent` and `parents` properties to all elements which allows parent traversal of an element in an element tree. We only support adding an element to a single tree and if the user tries to add the element to more than one element tree we will raise an exception.

- https://github.com/refractproject/minim/commit/4f59e7d62c2a50ad2a3f25d5617248f10c2d5278 Adds basically parent tracking
- https://github.com/refractproject/minim/commit/e2b056e01508559e58cacf74e1c4e58648313745 uses Proxy to observe content arrays for changes and update the `parent` property as necessary when a user tries to mutate an array set in `content` of an element.
- https://github.com/refractproject/minim/commit/2cab0094f576caa7be1ab57192da9b4b2eb3c45f adds a convince properties `parents` which returns all of the parents upwards for the element.
- https://github.com/refractproject/minim/commit/5e909f6cfa0497d4e206aad124e4b13bbaecaf4b removes the custom parent tracking in `findRecursive` in favour of the global element parent tracking.
- https://github.com/refractproject/minim/commit/12c3a63268a30717c429df37474cae3a3bcb6cb6 Unfortunately I discovered that `Proxy` is not available on Node 4 (without `--harmony-proxioes` node argument) so I have disabled its use. Please see commit message for all details. This is very problematic for Node 4 and introduces a tricky workaround.

Please review per-commit and see the commit messages for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Master
 
+## Breaking
+
 - `Element.children` and `Element.recursiveChildren` now return `ArraySlice`
   instead of an `ArrayElement`.
 - `ArrayElement.filter` and `ArrayElement.find*` now return `ArraySlice`
@@ -8,6 +10,11 @@
   instead of methods.
 - `ObjectElement.filter` now returns an `ObjectSlice` instead of an
   `ObjectElement`.
+
+## Enhancements
+
+- Every element in an element tree will now have a `parent` providing access to
+  traverse up an element tree from a child.
 
 # 0.18.1
 
--- a/README.md
+++ b/README.md
@@ -207,44 +207,6 @@ that we are recursively looking for all `string` elements that are found within 
 const stringInsideMembers = element.findRecursive('member', 'string');
 ```
 
-##### Parents
-
-Each returned element will include a new `parents` property which is an array
-element including the parents of the returned element.
-
-As an example, if I had an array element which contains a category element
-which in turn contains a string element with the content "Hello World". I can
-access the parent array and category element. The parents are in closest parent
-order.
-
-```json
-{
-  "element": "array",
-  "content": [
-    {
-      "element": "category",
-      "content": [
-        {
-          "element": "string",
-          "content": "Hello World"
-        }
-      ]
-    }
-  ]
-}
-```
-
-```javascript
-const elements = element.findRecursive('string');
-const helloString = elements.first;
-
-// Category Element
-helloString.parents[0];
-
-// Array Element
-helloString.parents[1];
-```
-
 #### children
 
 The `children` property returns an `ArrayElement` containing all of the direct children elements.
--- a/lib/key-value-pair.js
+++ b/lib/key-value-pair.js
@@ -7,6 +7,7 @@ var KeyValuePair = createClass({
   constructor: function(key, value) {
     this.key = key;
     this.value = value;
+    this.parent = null;
   },
 
   clone: function() {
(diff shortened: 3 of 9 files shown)


honzajavorek:
If I get it right, the last commit means [this line](https://github.com/apiaryio/dredd-transactions/blob/master/src/parse.coffee#L34) won't work on Node 4 anymore. It's the only place now where Dredd mutates the API Elements tree. Also, I guess similar operations happen a lot in the Swagger adapter, don't they?

Couldn't the content array be an object inheriting from standard JS array, which just overrides certain methods (just for parent tracking, then calling and returning from the "super" methods)? Maybe I'm imagining things impossible in the prototype inheritance, IDK. Sth along the lines of [this SO answer](https://stackoverflow.com/a/18188209/325365).

kylef:
> If I get it right, the last commit means this line won't work on Node 4 anymore

No thats not right, it will work. But it will be less optimal on Node 4 compared to Node 6. Since it will make a shallow copy of the array every time any mutable ArrayElement is called to discard the frozen version. This has a potential performance since we are resetting parent states for these elements. However, we are using Node 6 which provides Proxy support and thus this work around won't be used.

> Couldn't the content array be an object inheriting from standard JS array, which just overrides certain methods (just for parent tracking, then calling and returning from the "super" methods)?

I don't think it's possible override default methods on a builtin Array in ES5 or Node 4. I attempted it and it works in Node 8 but not Node 4.

How important is Node 4 support to us? I see a few options:

- Drop Node 4 support
    - Node 4 is only LTS maintenance. We could not provide new features in Node 4 and only provide stable patch releases to this LTS maintenance.
    - Node 6 is active LTS (long live Node 6)
- Support Node 4 using less performant mechanism (what I did in https://github.com/refractproject/minim/pull/137/commits/12c3a63268a30717c429df37474cae3a3bcb6cb6)
- Change implementation so that mutating the content array won't update parents (I'd rather not do this as it could lead to obscure bugs when other developers do this by accident/on purpose). Observing the array `content` is a much safer approach than only updating parent tracking when the array is mutated via our convinces in ArrayElement.
    - Since this isn't supported on the Node 4 support of this PR, the internal array is frozen so that any mutation would raise an exception (other than the mutation ArrayElement does itself via conveniences).
- Don't support parent tracking

I'm leaning to "Support Node 4 using less performant mechanism" if we can't drop Node 4.

honzajavorek:
Oh, then I misunderstood the outcome. I thought not only there would be
performance impact, but also it would be hard to mutate the tree now.

Now that's pretty smart what you've done and I agree it's the best option.
It will be no real performance impact for Dredd itself (not mutating, only
one place I linked and even that one is a special case). So the most
endangered would be Swagger users, but hey, this is still the best
solution. Node 4 users should be glad someone still thinks of them in means
of support.

honzajavorek:
BTW, I was able to implement the inheritance in CoffeeScript:

```coffee
class SpecialArray extends Array

  constructor: (args...) ->
    @_something = 0
    super(args...)

  push: (args...) ->
    @_something = 42
    super(args...)


sa = new SpecialArray()
sa.push(1)
sa.unshift(3)
sa.pop()

console.log(sa)
console.log(sa.join(', '))
console.log(sa._something)
```

Prints

```
SpecialArray { '0': 3, _something: 42, length: 1 }
3
42
```

I'm not really sure how much elegant that would be in plain JS.

kylef:
@honzajavorek Can you transpile and run that on Node 4? I did similar wit ES6 and Babel and the side-effects in Array subclass didn't run when involved with Node 4.

honzajavorek:
Hmm, good point. I tried on my current Node, which is 6 :( My excitement
from the experiment shadowed the original point - to make it work on 4.

honzajavorek:
I just left work, so can't try right away. I did not try to run it on 4. I
only tried to run on 6 and to see how the compiled JS looks like (looks
like... how to be polite... looks horrible).

honzajavorek:
```
$ coffee special.coffee
SpecialArray { '0': 3, _something: 42, length: 1 }
3
42

14:40 ~/Projects
$ coffee --version
CoffeeScript version 1.12.7

14:40 ~/Projects
$ node --version
v4.8.4
```

@kylef I can transpile and run it on Node 4.

kylef:
#141 contains an alternative solution to having `parents` which I think is a lot simpler as it doesn't handle "tracking" for changes. The PR adds a new `freeze` method which updates all of the parent properties and prevents the element from being mutated causing the parents to go out of sync. I think this should solve all of the same problems.

pksunkara:
@kylef I will leave it to you to close and delete this.

PR #140, opened 2017-08-02
Add jsdoc to public interfaces

kylef:
This PR adds JSDoc style comments to the public interfaces. I think this can allow us to provide far better documentation that we currently have in the readme and generate a website for all the documentation (minim, minim-api-description, minim-parse-result) eventually. Especially using [sphinx-js](https://hacks.mozilla.org/2017/07/introducing-sphinx-js-a-better-way-to-document-large-javascript-projects/). This pull request is the "first step" which is to add JSDoc comments, they're not perfect and could be iterated on in the future but I think this is a good start.

The jsdoc npm package can be used to generate a website locally, all of the types are hyperlinked.

```shell
$ npm install -g jsdoc
$ jsdoc -r lib
$ open out/index.html
```

Here's just a preview of one method:

screen shot 2017-08-02 at 16 48 06

### DIFF
--- a/lib/array-slice.js
+++ b/lib/array-slice.js
@@ -5,12 +5,22 @@ var uptown = require('uptown');
 var refract = require('./refraction').refract;
 var createClass = uptown.createClass;
 
+/**
+ * @class
+ *
+ * @param {Element[]} elements
+ *
+ * @property {Element[]} elements
+ */
 var ArraySlice = createClass({
-  // elements: [Element]
   constructor: function (elements) {
     this.elements = elements || [];
   },
 
+  /**
+   * @returns {Array}
+   * @memberof ArraySlice.prototype
+   */
   toValue: function () {
     return this.elements.map(function (element) {
       return element.toValue();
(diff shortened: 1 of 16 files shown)


lib/array-slice.js
```diff
@@ -19,24 +23,45 @@ var ArraySlice = createClass({
 
   // High Order Functions
 
+  /**
+   * @param {callback} - Function to execute for each element
+   * @param {thisArg} - Value to use as this (i.e the reference Object) when executing callback
+   * @memberof ArraySlice.prototype
```
kylef: Since we use ES5 which doesn't have classes, we have to add additional annotations like `@memberof`, `@readonly` etc. If we move to ES6 eventually (I could see this happening), we get to remove some of these as they can be inferred.

pksunkara: Yeah, we need to move minim to ES6

tjanc (dismissed):
Some functions are still to be documented:
- lib/array-slice.js: constructor, toValue
- lib/link-element.js: constructor
- lib/ref-element.js: constructor
- lib/key-value-pair.js: constructor, clone
- lib/object-slice.js: constructor, filter, forEach
- lib/primitives/*: constructor, primitive
- lib/primitives/array-element.js: getIndex, remove, findElements
- lib/primitives/element.js: equals, getMetaProperty, setMetaProperty
- lib/primitives/object-element.js: getKey

Also consider documenting getters and setters.

honzajavorek (approved):
LGTM - I think it's a great first step towards having better and consistent docs for minim. Not sure whether 100% docs coverage was an intent, but that's what @tjanc already pointed out.

lib/array-slice.js
smizell: Just to rehash why we did what we did, we used the uptown library I had to provide an interface for ES5 users to extend the functionality. Without it, they would need to manually handle the protoype creation, which is usually not straightforward. ES6 classes actually work with this, as you can extend with `class MyArraySlice extends ArraySlice` (unless something changed and it doesn't work anymore).

What could be done is to move to ES6 classes and then provide uptown either as a convenience function (like React and `createClass`) or just a note in the README stating how ES5 users can extend without copying everything manually to the prototype.

However, there are were regrets of moving React to ES6 classes and some reasons why Ember never did [here](https://twitter.com/wycats/status/892254920102952960). Not sure they apply, but something to consider.

smizell (approved):
I'm for it! Great idea.

kylef:
> Not sure whether 100% docs coverage was an intent, but that's what @tjanc already pointed out.

Yes, 100% perfect docs wasn't the intention. Just trying to get this started, I have made a lot of improvements since the review. I think this should be good to go as-is.

PR #141, opened 2017-08-03
Allow freezing and placing `parent` properties on elements

kylef:
This PR adds a new method on Element along with three new properties. You can now call the `freeze` method on an element which will prevent any mutations from happening to the element. Freezing an element will also add a `parent` accessor to all child elements allowing a user to traverse up the element tree.

Without having to support mutation, we can have a much simpler implementation compared to #137. The second commit in this PR will remove the custom parent tracking in findRecursive, since it is needed when you using `findRecursive` with multiple element names it will raise when the element is not frozen.

### DIFF
--- a/lib/primitives/element.js
+++ b/lib/primitives/element.js
@@ -143,52 +143,38 @@ var Element = createClass({
   },
 
   /**
-   * Finds the given elements in the element tree
+   * Finds the given elements in the element tree.
+   * When providing multiple element names, you must first freeze the element.
+   *
    * @param {...elementNames}
    * @returns {ArraySlice}
    * @memberof Element.prototype
    */
   findRecursive: function() {
-    var ArrayElement = require('./array-element');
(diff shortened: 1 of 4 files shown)


lib/primitives/element.js
```diff
@@ -30,6 +30,36 @@ var Element = createClass({
...
 
+  /**
+   * Freezes the element to prevent any mutation.
+   * A frozen element will add `parent` property to every child element
+   * to allow traversing up the element tree.
+   *
+   * @memberof Element.prototype
+   */
+  freeze: function() {
+    if (this._meta) {
```
kylef, line 41: Note, since `this.meta` getter has a side effect of creating an ObjectElement if it does not exist. This would lead into an infinite loop of creating meta on the meta of the meta of the meta.... so I am accessing the internal property directly.

pksunkara (approved):

PR #145, opened 2017-08-14
Fix some problems when deserialising enum elements in JSON 0.6 serialiser

kylef:
I found two bugs, commented below.

I've also refactored the tests so that there are separate tests for each feature and includes full branch coverage, before since there was a single test for all cases, some if/else cases were not covered.

### DIFF
--- a/lib/serialisers/json-0.6.js
+++ b/lib/serialisers/json-0.6.js
@@ -148,8 +148,13 @@ module.exports = createClass({
       element.attributes.remove('samples');
 
       if (samples) {
-        samples = samples.map(function(sample) {
-          return sample.get(0);
+        // Flatten samples array of array
+        var existingSamples = samples;
+        samples = new ArrayElement();
+        existingSamples.forEach(function(sample) {
+          sample.forEach(function (sample) {
+            samples.push(sample);
+          });
         });
 
         var sample = samples.shift();
@@ -161,9 +166,11 @@ module.exports = createClass({
         }
 
         element.attributes.set('samples', samples);
+      } else {
+        element.content = undefined;
       }
 
-      // Unwrap th3e default value
+      // Unwrap the default value
       var defaultValue = element.attributes.get('default');
       if (defaultValue) {
         defaultValue = defaultValue.get(0);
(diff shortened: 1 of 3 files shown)


lib/serialisers/json-0.6.js
```diff
@@ -143,8 +143,13 @@ module.exports = createClass({
       element.attributes.remove('samples');
 
       if (samples) {
-        samples = samples.map(function(sample) {
-          return sample.get(0);
```
kylef, line 152: This would discard some sample values, for example during this case in API Blueprint: https://github.com/apiaryio/drafter/blob/v3.2.7/test/fixtures/mson/enum-sample.ast.json#L138-L159

lib/serialisers/json-0.6.js
```diff
@@ -156,9 +161,11 @@ module.exports = createClass({
         }
 
         element.attributes.set('samples', samples);
+      } else {
+        element.content = undefined;
```
kylef, line 170: Without this, during this branch we would leave the enumerations content also inside content. Tested in `enum element deserialises content`. This results in the same elements being reused in same tree and means calling `freeze` on the parse result will throw exception. Luckily Dredd Transactions tests caught this problem.

lib/serialisers/json-0.6.js
```diff
@@ -143,8 +143,13 @@ module.exports = createClass({
...
       if (samples) {
-        samples = samples.map(function(sample) {
-          return sample.get(0);
+        // Flatten samples array of array
+        var existingSamples = samples;
+        samples = new ArrayElement();
+        existingSamples.forEach(function(sample) {
+          sample.forEach(function (sample) {
+            samples.push(sample);
+          });
```
pksunkara, line 157: We have lodash. We can use flatten from it here.

pksunkara (changes requested):
Just one minor comment.

lib/serialisers/json-0.6.js
kylef: That won't work because we have an Array Element of Array Element. The lodash flatten only supports `Array` unfortunately.

pksunkara (approved):

PR #146, opened 2017-08-14
Don't mutate when calling accessors while frozen

kylef:
Please read the changelog entry in PR to understand the reasons behind this change.

I think we can make it so that meta and attributes are no longer lazy loaded. This would be a larger architecture change to do so and would be breaking so I will not do it during this fix. We will have to discuss this further as meta/attributes couldn't be ObjectElement themselves because then they would have meta/attributes to populate this making the constructor for an element cause an infinite loop.

### DIFF
--- a/test/primitives/base-element-test.js
+++ b/test/primitives/base-element-test.js
@@ -669,6 +669,19 @@ describe('Element', function() {
         element.attributes.set('key', 'value');
       }).to.throw();
     });
+
+    it('returns frozen objects from meta/attributes accessors', function () {
+      // An elements meta and attributes are lazy loaded and created on access.
+      // This would cause problems because that means you cannot access
+      // meta/attributes on frozen elements because the accessor has
+      // side-effects of creation.
+
+      var element = new minim.Element();
+      element.freeze();
+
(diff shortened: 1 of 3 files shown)


test/primitives/base-element-test.js
```diff
@@ -669,6 +669,19 @@ describe('Element', function() {
...
+    it('returns frozen objects from meta/attributes accessors', function () {
+      // An elements meta and attributes are lazy loaded and created on access.
+      // This would cause problems because that means you cannot access
+      // meta/attributes on frozen elements because the accessor has
+      // side-effects of creation.
+
+      var element = new minim.Element();
+      element.freeze();
+
+      expect(element.meta.isFrozen).to.be.true;
```
kylef, line 682: This would fail without the fix:

```
  1) Element freezing an element returns frozen objects from meta/attributes accessors:
     TypeError: Cannot add property _meta, object is not extensible
      at constructor.get (lib/primitives/element.js:271:20)
      at Context.<anonymous> (test/primitives/base-element-test.js:677:21)
```

pksunkara (approved):

PR #149, opened 2017-08-24
Add find and flatMap capabilities to ArraySlice

kylef:
See the individual commits for further information. This should provide all the popular high order functions on an ArraySlice and allow simplification of handling slices and simplify https://github.com/apiaryio/dredd-transactions/pull/101.

This allows the following behaviours:

#### Finding specific element by name from parent, children or recursiveChildren tree:

```js
element.parents.find('parseResult')
element.children.find('httpRequest')
element.recursiveChildren.find('httpRequest')
```

#### Get all of the title values of children (discard when no title)

```js
const titles = element.children.flatMap(element => element.meta.getValue('title'))
```

### DIFF
--- a/lib/array-slice.js
+++ b/lib/array-slice.js
@@ -60,6 +60,27 @@ var ArraySlice = createClass({
   },
 
   /**
+   * Maps and then flattens the results.
+   * @param callback - Function to execute for each element.
+   * @param thisArg - Value to use as this (i.e the reference Object) when executing callback
+   * @returns {Element}
+   * @memberof ArraySlice.prototype
+   */
+  flatMap: function (callback, thisArg) {
+    var results = [];
+
+    this.forEach(function (element) {
+      var result = callback(element);
+
+      if (result) {
+        results.push(result);
+      }
+    }, this);
+
+    return results;
+  },
+
+  /**
    * @param callback - Function to execute for each element. This may be a callback, an element name or an element class.
    * @param thisArg - Value to use as this (i.e the reference Object) when executing callback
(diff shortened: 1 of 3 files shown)


lib/array-slice.js
```diff
@@ -40,16 +61,50 @@ var ArraySlice = createClass({
...
+   * @param callback - Function to execute for each element.
+   * @param thisArg - Value to use as this (i.e the reference Object) when executing callback
+   * @returns {Element}
+   * @memberof ArraySlice.prototype
+   */
+  flatMap: function (callback, thisArg) {
+    var results = [];
+
+    this.forEach(function (element) {
+      var result = callback(element);
```
honzajavorek, line 73: The `thisArg` is unused.

```diff
- callback(element)
+ callback.call(thisArg, element)
```

lib/array-slice.js
```diff
@@ -40,16 +61,50 @@ var ArraySlice = createClass({
...
+
+      if (result) {
+        results.push(result);
+      }
+    }, this);
```
honzajavorek: The `this` here is probably redundant - the `forEach` callback does not access `this`, so it doesn't need to be changed.

honzajavorek (dismissed):
All good, except `thisArg` is unused in the `flatMap` function.

lib/array-slice.js
kylef: I'll update this to be `thisArg` I think that was my intention and I mistakenly put `this`. Would solve the above comment also because then the callback will have this correctly.

pksunkara (approved):

PR #153, opened 2017-11-01
Add Node 8 to the Travis tests.

realityking:


kylef (approved):

kylef:
Thanks @realityking, we should definitely be testing here as Node 8 is now LTS.

PR #155, opened 2017-11-22
feat: add reject operator

char0n:
Implemented as complement of filter operator


pksunkara (approved):

honzajavorek:
@char0n Isn't `reject` way to close to naming common when using promises? It could be confusing for someone. When I read `reject`, I imagine promises at once.

char0n:
Not actually, it's a common naming in FP for complement of filtering the list like structures.
http://ramdajs.com/docs/#reject

PR #156, opened 2017-11-24
Add changelog entry for reject

kylef:
See https://github.com/refractproject/minim/pull/155, wanted to add this so it doesn't get forgotten.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Master
 
+## Enhancements
+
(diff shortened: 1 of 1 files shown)


honzajavorek:
Isn't `reject` way to close to naming common when using promises? It could be confusing for someone. When I read `reject`, I imagine promises at once.

pksunkara:
Can you rebase over master?

pksunkara (approved):

PR #157, opened 2017-11-27
Update enum deserialiser for changes in API Elements

kylef:


lib/serialisers/json-0.6.js
```diff
@@ -24,21 +24,23 @@ module.exports = createClass({
...
-      } else {
-        var attributes = element.attributes;
-
-        // Meta attribute was renamed to metadata
-        if (attributes.get('metadata')) {
-          attributes = attributes.clone();
-          attributes.set('meta', attributes.get('metadata'));
-          attributes.remove('metadata');
-        }
+    if (this[element.element + 'SerialiseAttributes']) {
```
kylef, line 27: Since the custom `SerialiseAttributes` handler may add attributes even if there we're non already we shouldn't check for `element.attributes.length > 0` first.

lib/serialisers/json-0.6.js
```diff
@@ -151,19 +162,23 @@ module.exports = createClass({
...
-        // Flatten samples array of array
+        // Re-wrap samples from array of array to array of enum's
+
         var existingSamples = samples;
+
         samples = new ArrayElement();
         existingSamples.forEach(function(sample) {
           sample.forEach(function (sample) {
-            samples.push(sample);
+            var enumElement = new ElementClass(sample);
```
pksunkara, line 172: I guess, we can use a new `EnumElement` that we talked about.

pksunkara (approved):

PR #158, opened 2017-11-29
feat: Added EnumElement

pksunkara:


test/serialisers/json-0.6.js
```diff
@@ -176,20 +176,18 @@ describe('JSON Serialiser', function() {
       var defaultElement = new minim.Element(new minim.elements.String('North'));
       defaultElement.element = 'default';
 
-      var sampleNorth = new minim.Element(new minim.elements.String('North'));
-      sampleNorth.element = 'enum';
-      var sampleEast = new minim.Element(new minim.elements.String('East'));
-      sampleEast.element = 'enum';
+      var sampleNorth = new minim.elements.Enum(new minim.elements.String('North'));
```
kylef, line 179: As the content gets refracted (`Element.call(this, refract(content), meta, attributes);)` this could be simplified as:

```js
var sampleNorth = new minim.elements.Enum('North');
```

kylef (commented):
Since `enum` is defined in API Elements, shouldn't it be in the other minim packages?

---

Although, to be completely honest with you I'd like change the structure of the packages (outside of this change) in the future.

PR #164, opened 2018-01-05
Don't create meta/attributes when checking if empty

kylef:
The `meta` and `attributes` properties of an element are computed, in the case that meta and attributes are not set when access. An empty `ObjectElement` is returned.

When we check if meta or attributes is empty, if we access via the getter there may be an unnecessary instantiation of an `ObjectElement` which slows down serialisation when meta or attributes is not set.

### DIFF
--- a/lib/serialisers/json-0.6.js
+++ b/lib/serialisers/json-0.6.js
@@ -20,17 +20,17 @@ module.exports = createClass({
       element: element.element,
     };
 
-    if (element.meta.length > 0) {
+    if (element._meta && element._meta.length > 0) {
       payload['meta'] = this.serialiseObject(element.meta);
     }
 
-    if (this[element.element + 'SerialiseAttributes']) {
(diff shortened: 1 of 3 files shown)


lib/serialisers/json-0.6.js
```diff
@@ -20,17 +20,17 @@ module.exports = createClass({
...
     };
 
-    if (element.meta.length > 0) {
+    if (element._meta && element._meta.length > 0) {
       payload['meta'] = this.serialiseObject(element.meta);
     }
 
-    if (this[element.element + 'SerialiseAttributes']) {
-      var attributes = this[element.element + 'SerialiseAttributes'](element);
+    if (element.element === 'enum') {
```
kylef, line 27: Since there is only custom serialise attributes for enum, it's faster to explicitly check instead of dynamically dispatch.

pksunkara (approved):

pksunkara:
Approved PR. But please fix the failing tests before merge.

PR #166, opened 2018-01-09
Use native ES6 classes over Uptown

kylef:
This removes the dependency on Uptown and uses ES6 classes which are available in Node 4, the lowest Node we support.

Transpilation is done for the browser distribution to maintain compatibility for all browsers.

I've also added use of const/let over var while I was at it.

#### DO NOT MERGE

As of right now, this change would break downstream minim-api-description and minim-parse-result due to https://github.com/babel/babel/issues/6299 (Babel-transpiled class can't extend from native class), a fix for this can be found in Babel 7 (there are betas available but no final release). We could move peasant and other minim libraries to Babel 7.

An alternative, perhaps we can move minim-* packages to work similar to minim. Where the node sources are not transpiled and instead written for plain Node 4 and then just transpired for the web distribution.


pksunkara:
I would prefer the babel7 thing. Otherwise, we have to change a lot of code

pksunkara:
@kylef Do we need this anymore?

kylef:
I'd love to move to ES6 at some point, but I'd say it isn't a priority (if it isn't broken don't "fix" it). Will close PR as it certainly needs to be done again as things have changed and there is large conflict which will be difficult to fix. If done again we should do it more incrementally piece by piece.

PR #172, opened 2018-03-16
fix: Serialise enum with content to 0.6 correctly

pksunkara:
I am not terribly sure the fix is correct here. The document which is responsible for crashing is:

```
swagger: '2.0'
info:
  title: Crash
  version: 1
paths:
  /:
    get:
      parameters:
        - name: interests
          type: array
          items:
            type: string
          enum:
            - Homeless
          in: query  
          description: Interest
      responses:
        '401':
          description: Unauthorised
```

You could reproduce the crash by doing the following:

```
fury crash.yaml -f 'application/vnd.refract.parse-result+json; version=0.6'
```

### DIFF
--- a/lib/serialisers/json-0.6.js
+++ b/lib/serialisers/json-0.6.js
@@ -120,7 +120,7 @@ module.exports = createClass({
     }
 
     if (element.content) {
-      return [this.serialise(element.content)];
+      return this.serialiseContent(element.content);
     }
 
     return [];
--- a/test/serialisers/json-0.6.js
+++ b/test/serialisers/json-0.6.js
@@ -3,13 +3,13 @@ var expect = require('../spec-helper').expect;
 var Namespace = require('../../lib/minim').Namespace;
 var minim = require('../../lib/minim').namespace();
 var KeyValuePair = require('../../lib/key-value-pair');
-var JSONSerialiser = require('../../lib/serialisers/json-0.6');
+var JSON06Serialiser = require('../../lib/serialisers/json-0.6');
 
 describe('JSON 0.6 Serialiser', function() {
   var serialiser;
 
   beforeEach(function () {
-    serialiser = new JSONSerialiser(minim);
+    serialiser = new JSON06Serialiser(minim);
   });
 
   describe('initialisation', function() {
@@ -18,7 +18,7 @@ describe('JSON 0.6 Serialiser', function() {
     });
 
     it('creates a default namespace when no namespace is given', function() {
-      serialiser = new JSONSerialiser();
+      serialiser = new JSON06Serialiser();
       expect(serialiser.namespace).to.be.instanceof(Namespace);
     });
   });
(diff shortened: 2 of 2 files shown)


lib/serialisers/json-0.6.js
```diff
@@ -120,7 +120,7 @@ module.exports = createClass({
     }
 
     if (element.content) {
-      return [this.serialise(element.content)];
+      return this.serialiseContent(element.content);
```
pksunkara, line 123: This might be `[this.serialiseContent(element.content)]`. But I am not confident in either of them.

test/serialisers/json-0.6.js
```diff
@@ -285,6 +285,40 @@ describe('JSON 0.6 Serialiser', function() {
       });
     });
 
+    it('serialises enum with content', function () {
+      var enumeration = new minim.Element();
+      enumeration.element = 'enum';
+      enumeration.attributes.set('enumerations', []);
+      enumeration.content = [new minim.elements.String(null)];
```
kylef, line 292: The enumerations content (value) shouldn't be an array per the spec. It should be an element itself.

messages image 2872221322

lib/serialisers/json-0.6.js
kylef: We should be wrapping the content in an array, because in API Elements 1.0 (the `Element` object), content should be an element. In API Elements 0.6 (what we are serialising to), the content should be an array. Thus the content should be wrapped in an array.

kylef:
I think we can close this in favour of https://github.com/apiaryio/fury-adapter-swagger/pull/176

The following result when my patch is applied:

```
$ node lib/fury.js crash.yaml -f 'application/vnd.refract.parse-result+json; version=0.6'
{
  "element": "parseResult",
  "content": [
    {
      "element": "category",
      "meta": {
        "classes": [
          "api"
        ],
        "title": "Crash"
      },
      "content": [
        {
          "element": "resource",
          "attributes": {
            "href": "/"
          },
          "content": [
            {
              "element": "transition",
              "attributes": {
                "href": "/{?interests}",
                "hrefVariables": {
                  "element": "hrefVariables",
                  "content": [
                    {
                      "element": "member",
                      "meta": {
                        "description": "Interest"
                      },
                      "content": {
                        "key": {
                          "element": "string",
                          "content": "interests"
                        },
                        "value": {
                          "element": "array",
                          "content": [
                            {
                              "element": "string",
                              "content": null
                            }
                          ]
                        }
                      }
                    }
                  ]
                }
              },
              "content": [
                {
                  "element": "httpTransaction",
                  "content": [
                    {
                      "element": "httpRequest",
                      "attributes": {
                        "method": "GET"
                      }
                    },
                    {
                      "element": "httpResponse",
                      "attributes": {
                        "statusCode": "401"
                      },
                      "content": [
                        {
                          "element": "copy",
                          "content": "Unauthorised"
                        }
                      ]
                    }
                  ]
                }
              ]
            }
          ]
        }
      ]
    },
    {
      "element": "annotation",
      "meta": {
        "classes": [
          "warning"
        ],
        "links": [
          {
            "element": "link",
            "attributes": {
              "relation": "origin",
              "href": "http://docs.apiary.io/validations/swagger#swagger-validation"
            }
          }
        ]
      },
      "attributes": {
        "code": 6,
        "sourceMap": [
          {
            "element": "sourceMap",
            "content": [
              [
                212,
                8
              ]
            ]
          }
        ]
      },
      "content": "Expected type array but found type string"
    }
  ]
}
warning: (6)  Expected type array but found type string
```

PR #174, opened 2018-04-24
Strip fixed typeAttribute for enumerations

kylef:
The `fixed` typeAttribute was never used in enumerations in existing JSON 0.6 and is introduced in newer enum element found in API Elements 1.0.

### DIFF
--- a/lib/serialisers/json-0.6.js
+++ b/lib/serialisers/json-0.6.js
@@ -82,10 +82,14 @@ module.exports = createClass({
     // Wrap default in array
     var defaultValue = attributes.get('default');
     if (defaultValue && defaultValue.content) {
+      defaultValue.content.attributes.remove('typeAttributes');
       attributes.set('default', new ArrayElement([defaultValue.content]));
     }
 
     var samples = attributes.get('samples') || new ArrayElement([]);
+    samples.forEach(function (sample) {
+      sample.content.attributes.remove('typeAttributes');
+    });
 
     // Content -> Samples
     if (element.content && enumerations.length !== 0) {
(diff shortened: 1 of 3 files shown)


test/serialisers/json-0.6.js
```diff
@@ -185,7 +185,7 @@ describe('JSON 0.6 Serialiser', function() {
 
     it('serialises enum', function() {
       var defaultElement = new minim.Element(new minim.elements.String('North'));
-      defaultElement.element = 'default';
+      defaultElement.element = 'enum';
```
kylef, line 188: I spotted this was incorrect while copying the test case below to make a fixed variant.

pksunkara: Yup, this is what that led me into the whole default thing research.

lib/serialisers/json-0.6.js
```diff
@@ -82,10 +82,14 @@ module.exports = createClass({
     // Wrap default in array
     var defaultValue = attributes.get('default');
     if (defaultValue && defaultValue.content) {
+      defaultValue.content.attributes.remove('typeAttributes');
```
pksunkara, line 85: Are you sure you want to remove the whole of typeAttributes instead of just fixed?

kylef: You'd never have required, optional or any other typeAttribute in this place in 0.6? I was thinking perhaps it wasn't worth the additional logic as it shouldn't happen. But we could alter it if you think it might happen.

pksunkara: Hmm.. I've thought long and hard about this. I am proposing to leave it as it is.

pksunkara (approved):

PR #175, opened 2018-05-10
ArraySlice.flatMap and ArraySlice.reduce test cases

abacaphiliac:
Adds ArraySlice test cases copied from [Array.reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce#Examples) and [Array.flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Examples).

ArraySlice.flatMap had to be adjusted slightly to flatten an array of arrays, per Mozilla's test cases. That change is compatible with the existing ArraySlice.flatMap test. Are we OK with that difference in behavior between ArraySlice.flatMap and Array.flatMap? Is aligning with Array.flatMap desirable?

### DIFF
--- a/lib/array-slice.js
+++ b/lib/array-slice.js
@@ -73,7 +73,9 @@ var ArraySlice = createClass({
     this.forEach(function (element) {
       var result = callback(element);
 
-      if (result) {
+      if (Array.isArray(result)) {
+        results.push.apply(results, result);
+      } else if (result) {
         results.push(result);
       }
     }, thisArg);
--- a/test/array-slice-test.js
+++ b/test/array-slice-test.js
@@ -214,6 +214,53 @@ describe('ArraySlice', function () {
     expect(flattened).to.deep.equal([0, 1, 2, 3, 4, 5]);
   });
 
+  /**
+   * @link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#Alternative
+   */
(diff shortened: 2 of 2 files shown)


kylef:
> Is aligning with Array.flatMap desirable?

👍 I think it makes sense to align to how the standard Array `flatMap` implementation works.

---

Looks like the Mozilla code is permitted without license as it is public domain, so from legal perspective this is fine in terms of using those tests (just mentioning this so that others looking at PR are aware):

> No licensing notice is necessary, but if you need one, you can use: "Any copyright is dedicated to the Public Domain. http://creativecommons.org/publicdomain/zero/1.0/".

kylef (approved):

kylef:
I'm approving the changes, I'm going to hold off merging right now as it would be a breaking change to the interface (if there is anyone whom is relying on the existing behaviour).

I'll need to figure out which release this will end up in and the release coordination (since minim, fury and the adapters are currently tied to very specific versions).

abacaphiliac:
in the spirit of aligning to Array.flatMap, should i drop the [original source line](https://github.com/refractproject/minim/pull/175/files#diff-272b6f48699f6d7faf9b46ee15e3da9aR79) and update the existing flatMap test?

pksunkara (approved):

pksunkara:
This should end up in v0.21.0

PR #183, opened 2018-07-23
feat: provide isMinim for ducktyping

opichals:
Would help [here](https://github.com/refractproject/minim-api-description/issues/38) to create a generic `element` predicate while avoiding the use of the generally discouraged `instanceof` operator.

### DIFF
--- a/lib/array-slice.js
+++ b/lib/array-slice.js
@@ -39,6 +39,12 @@ var ArraySlice = createClass({
   },
 
   /**
+   * Ducktyping support (to avoid using instanceof)
+   * @memberof ArraySlice.prototype
+   */
+  isMinim: true,
+
+  /**
    * @returns {Array}
    * @memberof ArraySlice.prototype
    */
--- a/lib/primitives/element.js
+++ b/lib/primitives/element.js
@@ -40,6 +40,12 @@ var Element = createClass({
   },
 
   /**
+   * Ducktyping support (to avoid using instanceof)
+   * @memberof Element.prototype
+   */
+  isMinim: true,
+
+  /**
    * Freezes the element to prevent any mutation.
    * A frozen element will add `parent` property to every child element
    * to allow traversing up the element tree.
--- a/test/array-slice-test.js
+++ b/test/array-slice-test.js
@@ -5,6 +5,12 @@ var StringElement = minim.StringElement;
 var ArraySlice = minim.ArraySlice;
 
 describe('ArraySlice', function () {
+  it('supports isMinim ducktyping check', function () {
+    var slice = new ArraySlice([new Element()]);
+
+    expect(slice.isMinim).to.be.true;
+  });
+
   it('can be created from an array of elements', function () {
     var element = new Element();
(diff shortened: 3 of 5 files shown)


opichals:
Our current documentation workaround is to check `isFunction(element.toValue)` as we found the `isFunction(element.getMetaProperty)` fails for `ArraySlice`.

honzajavorek:
I don’t want to pick on words, but isn’t check on .toValue something one would call duck typing?

I mean, I thought duck typing is assuming something has certain behavior and if it doesn’t, then handle it differently/raise error.

Checking .isMinim sounds like an opposite. It doesn’t use instanceof, but it’s IMHO the same approach - type checking before I do something, only using a custom property unrelated to the behavior itself instead of inheritance built in the language.

opichals:
@honzajavorek Yes, perhaps checking the `toValue` is just enough and is what ducktyping is about.

For me it was about the likelihood of such a function to be present on the object which is subjectively higher to me than in the `isMinim`'s case.

opichals:
@kylef Could you please have a look?

kylef:
@opichals Can you please exlain why using `instanceof` is discouraged?

opichals:
@kylef I could mainly enumerate the following
* instanceof minim.Element doesn't work as the sole minim check (`minim.ArraySlice instanceof minim.Element === false`)
* when minim is bundled along with a library (e.g. attributes-kit) the instanceof checks fail when the minim.Element instances are created in the application scope

Also there is an explanation of the issue in the [minim predicates module README](https://github.com/apiaryio/documentation-service/blob/master/phoenix-ui/src/documentation/utils/minim/predicates/README.md#justification)

kylef (changes requested):
@opichals Ok, I think it would make more sense to change the name, and also to differentiate between slice and element.

I don't want to couple the name of the library into its public API, in the future we intend for consumers to use [api-elements.js](https://github.com/apiaryio/api-elements.js) and they shouldn't have to know about implementation details like the underlying package is called minim etc. Having properties such as `isMinim` would expose these kinds of implementation details.

I think having `isElement` and `isElementSlice` properties would be clearer, and allow you to actually know if its an element or a slice.

kylef:
> When creating a new minim namespace the new types/classes are created in runtime and there is no way to statically use instanceof to assert the proper type. The predicates are only concerned by the shape of the elements not by an Is-a relations which is the tightest form of coupling and should be generally avoided.

@opichals I'd dispute that reasoning for minim to not use `instanceof`. Base element types are not created in runtime and thus it would be possible to use `instanceof`.

opichals:
> Base element types are not created in runtime and thus it would be possible to use instanceof.

/cc @char0n

opichals:
> I think having isElement and isElementSlice properties would be clearer, and allow you to actually know if its an element or a slice.

To me the `isElement` wording is rising the likelihood of a naming collision.

[Used instanceof in attributes-kit](https://github.com/apiaryio/attributes-kit/blob/master/src/Components/Attributes/Attributes.js#L27) so closing this.

PR #185, opened 2018-08-29
Deserialise meta into metadata in 0.6 deserialiser

kylef:
This fixes the bug that is worked around in https://github.com/apiaryio/Paw-APIBlueprintImporter/blob/568d73e44bdff23495c24fb3f4b5cd2d1cc8025a/src/APIElementImporter.js#L8 and is also causing documentation team problems.

Meta will be correct deserialised in 0.6 as metadata as an object element with this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# Master
+
+## Bug Fixes
+
+- JSON 0.6 deserialiser will now correct deserialise an API Categories `meta`
+  attribute into `metadata`.
+
 # 0.20.5
 
 ## Bug Fixes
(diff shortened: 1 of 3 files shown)


pksunkara (approved):

opichals:
@kylef after upgrading we see a change the `meta.toValue()` contents:

minim@0.20.5 element.attributes.get('meta').toValue():
```
[
  {
    "key": "FORMAT",
    "value": "1A"
  },
  {
    "key": "HOST",
    "value": "https://localhost/"
  }
]
```

minim@0.20.6 element.attributes.get('metadata').toValue():
```
{
  "FORMAT": "1A",
  "HOST": "https://localhost/"
}
```

PR #189, opened 2018-09-03
Reduce use of lodash & load only those parts that are used

realityking:
This mostly makes it a lot easier to see what parts of lodash are actually used by minim. Turns out it's only 8 methods.

### DIFF
--- a/lib/array-slice.js
+++ b/lib/array-slice.js
@@ -3,7 +3,7 @@
 var uptown = require('uptown');
 var refract = require('./refraction').refract;
 var createClass = uptown.createClass;
-var _ = require('lodash');
+var negate = require('lodash/negate');
 
 // Coerces an a parameter into a callback for matching elements.
 // This accepts an element name, an element type and returns a
@@ -100,7 +100,7 @@ var ArraySlice = createClass({
    */
   reject: function (callback, thisArg) {
     callback = coerceElementMatchingCallback(callback);
-    return new ArraySlice(this.elements.filter(_.negate(callback), thisArg));
+    return new ArraySlice(this.elements.filter(negate(callback), thisArg));
   },
 
   /**
--- a/lib/namespace.js
+++ b/lib/namespace.js
@@ -1,6 +1,10 @@
 'use strict';
 
-var _ = require('lodash');
+var isNull = require('lodash/isNull');
+var isString = require('lodash/isString');
+var isNumber = require('lodash/isNumber');
+var isBoolean = require('lodash/isBoolean');
+var isObject = require('lodash/isObject');
 var uptown = require('uptown');
 var createClass = uptown.createClass;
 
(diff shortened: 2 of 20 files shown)


kylef:
Since uptown (one of the depedencies) brings its own require for lodash. This has a negative impact on the dist sizes and makes them larger (you would think it would make them smaller but it just includes more of lodash instead).

I've made a PR to uptown (https://github.com/smizell/uptown/pull/8) which would do a similiar change in uptown to the one in this PR.

To demonstate, I have taken the patch for package.json so that lodash is included in bundle and generated the dist from master:

```
$ du -h dist/*
1.5M    dist/minim.js <- master with lodash
1.6M    dist/minim.js <- this branch
````

You can see depending on the sub packages for lodash results in larger dist sizes overall. I've then linked to the branch of uptown in https://github.com/smizell/uptown/pull/8 and then created the bundles, you can see it has huge impact then:

```
$ du -h dist/*
416K    dist/minim.js
 88K    dist/minim.min.js
````

NOTE, added the minified version like uptown:

```diff
diff --git a/package.json b/package.json
index 68a2190..3a93ce8 100644
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "main": "lib/minim.js",
   "scripts": {
     "prepare": "npm run build",
-    "build": "browserify -d -s minim -o dist/minim.js lib/minim.js",
+    "build": "npm run build:full && npm run build:min",
+    "build:full": "browserify -d -s minim -o dist/minim.js lib/minim.js",
+    "build:min": "cat dist/minim.js | uglifyjs > dist/minim.min.js",
     "coverage": "istanbul cover _mocha -- -R spec --recursive",
     "coveralls": "coveralls <coverage/lcov.info",
     "frontendtests": "karma start",
@@ -47,7 +49,8 @@
     "mocha-sinon": "^2.1.0",
     "sinon": "^5.0.10",
     "sinon-chai": "^3.1.0",
-    "watchify": "^3.11.0"
+    "watchify": "^3.11.0",
+    "uglify-js": "^3.2.2"
   },
   "dependencies": {
     "lodash": "^4.15.0",
```

So I'd get https://github.com/smizell/uptown/pull/8 merged and released before and then we can take this PR with the uptown update! Thanks for this @realityking.

/c @char0n Does including lodash in the browser bundle effect/impact you with frontend documentation? Perhaps the changing the requires may so I think I'd like your review (or you can delegate to someone on your team, I don't want to break things downstream)

realityking:
Oh my bad, I hadn't considered browser bundling.

opichals:
@kylef Thanks! The frontend documentation would only benefit from this. It should not break anything there and if it does the best course of action is to fix it to work with this PR merged.

char0n:
Yes I confirm, we will benefit from it. FYI there is a webpack plugin https://github.com/lodash/lodash-webpack-plugin that tries to solve this problem in the same way.

kylef:
@realityking uptown 1.1.0 has been updated, would you be able to bump that version inside package.json and we can merge this PR.

kylef:
@realityking and thanks for bringing this up. Saves a lot of space in the resultant browser bundles.

realityking:
@kylef Sorry this took so long but I pushed the change :)

kylef (approved):

kylef:
Thanks @realityking

PR #190, opened 2018-09-24
Add flatMap to Array and Object element

kylef:
We already provide flatMap on ArraySlice but it isn't available on the base elements which this PR adds.

### DIFF
--- a/lib/array-slice.js
+++ b/lib/array-slice.js
@@ -67,10 +67,21 @@ var ArraySlice = createClass({
    * @memberof ArraySlice.prototype
    */
   flatMap: function (callback, thisArg) {
+    return this.compactMap(callback, thisArg);
+  },
+
+  /**
+   * Returns an array containing the truthy results of calling the given transformation with each element of this sequence
+   * @param transform - A closure that accepts an element of this array as its argument and returns an optional value.
+   * @param thisArg - Value to use as this (i.e the reference Object) when executing callback
+   * @memberof ArraySlice.prototype
+   * @returns An array of the non-undefined results of calling transform with each element of the array
+   */
+  compactMap: function (transform, thisArg) {
     var results = [];
 
     this.forEach(function (element) {
-      var result = callback(element);
+      var result = transform(element);
 
       if (result) {
         results.push(result);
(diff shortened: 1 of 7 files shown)


pksunkara:
What's the reason behind the name `flat`? It means something else.

```
var arr1 = [1, 2, [3, 4]];
arr1.flat(); 
// [1, 2, 3, 4]
```

What you are doing is removing the undefined null values. There should be another name for that.

pksunkara (dismissed):
Just minor name change

kylef:
@pksunkara, I've renamed the method to `compactMap` which is similiary named to the Swift equivilent (that also used to be called `flatMap` (https://github.com/apple/swift-evolution/blob/571347649429cb08e73a2bb6feabe6ff2428db73/proposals/0187-introduce-filtermap.md)).

lib/array-slice.js
```diff
@@ -65,13 +65,25 @@ var ArraySlice = createClass({
    * @param callback - Function to execute for each element.
    * @param thisArg - Value to use as this (i.e the reference Object) when executing callback
    * @returns {Element}
+   * @deprecated
    * @memberof ArraySlice.prototype
    */
   flatMap: function (callback, thisArg) {
+    return this.compactMap(callback, thisArg);
```
pksunkara, line 70: We should remove this completely, since we are still pre 1.0, we can afford to break it and need no deprecation periods.

pksunkara (changes requested):

lib/array-slice.js
kylef: I want to remove it in the next minor release, but given how hard it is to make a minor release (we have to release entire minim and fury toolchain for breaking changes) then I would put this in next patch release of Minim and remove the deprecated method in the next minor release (when that happens).

I'd wait with minor release until after the differentiator is ran incase we have further fixes to make to the JSON 0.6 logic.

kylef: Given we have https://github.com/refractproject/minim/pull/175 open, I'd suggest we merge this PR as-is and then I'll rebase #175 and make those tests pass. So `flatMap` behaves like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap. Since current behaviour is something else, modelled from `flatMap` from other languages.

pksunkara (approved):

PR #194, opened 2018-11-02
More tests and fixes to 0.6 serialization

pksunkara:


lib/serialisers/json-0.6.js
```diff
@@ -16,7 +16,9 @@ module.exports = JSONSerialiser.extend({
       payload['meta'] = this.serialiseObject(element.meta);
     }
 
-    if (element.element === 'enum') {
+    var isEnum = (element.element === 'enum' || element.attributes.keys().indexOf('enumerations') !== -1);
```
kylef, line 19: Why would we have `enumerations` that isn't in an `enum` element?

lib/serialisers/json-0.6.js
```diff
@@ -317,7 +323,14 @@ module.exports = JSONSerialiser.extend({
...
       // additional information
-      // TODO
+      var values = [];
+      var content = item.content || [];
+
+      for (var index = 0; index < content.length; index++) {
+        values.push(this.serialise(content[index]));
+      }
+
+      return values;
```
kylef, line 333: This could be simpler using map:

```js
return (item.content || []).map(this.serialise);
```

Possibly needs thisArg:

```js
return (item.content || []).map(this.serialise, this);
```

kylef:
Two comments, otherwise looks good.

lib/serialisers/json-0.6.js
pksunkara: When a data structure is inheriting from an Enum named type. The element name would be the name of the type but it can contain enumerations

lib/serialisers/json-0.6.js
pksunkara: I want to leave this as it is, because it might need some modifications later on. I am not sure this fixes all the cases that can arise.

kylef (approved):

PR #196, opened 2018-11-06
Don't serialise empty arrays unless request/response or category in JSON 0.6

kylef:
This aligns with Drafter 3 behaviour where empty arrays are not serialised, except for request/response or category elements.

### DIFF
--- a/lib/serialisers/json-0.6.js
+++ b/lib/serialisers/json-0.6.js
@@ -64,12 +64,28 @@ module.exports = JSONSerialiser.extend({
         content = this.serialiseContent(element.content);
       }
 
-      if (content !== undefined) {
+      if (this.shouldSerialiseContent(element, content)) {
         payload['content'] = content;
       }
     }
 
     return payload;
+  },
+
+  shouldSerialiseContent: function(element, content) {
+    if (content === undefined) {
+      return false;
+    }
+
+    if (element.element === 'httpRequest' || element.element === 'httpResponse' || element.element === 'category') {
+      return true;
+    }
+
+    if (Array.isArray(content) && content.length === 0) {
+      return false;
+    }
+
+    return true;
   },
 
   refSerialiseContent: function(element, payload) {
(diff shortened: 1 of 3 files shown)


lib/serialisers/json-0.6.js
```diff
@@ -44,14 +44,30 @@ module.exports = JSONSerialiser.extend({
...
 
     return payload;
   },
 
+  shouldSerialiseContent: function(element, content) {
+    if (content === undefined) {
+      return false;
+    }
+
+    if (element.element === 'httpRequest' || element.element === 'httpResponse' || element.element === 'category') {
```
pksunkara, line 80: I don't like this line. It's minim-parse-result specific.

pksunkara (changes requested):

lib/serialisers/json-0.6.js
kylef: We already have various logic here that is minim-api-description and minim-parse-result-specific. This is due to Refract 0.6 having element specific serialisation rules which we are moving away from in 1.0.

The whole point of this file is to bridge the gap from the old state to the future(current).

> ```js
> sourceMapSerialiseContent: function(element) {
>   return element.toValue();
> },
> ```

> ```js
> // "meta" attribute has been renamed to metadata
> var metadata = element.attributes.get('meta');
> ```

> ```js
> dataStructureSerialiseContent: function(element) {
>   return [this.serialiseContent(element.content)];
> },
> ```

honzajavorek: Could the line be more readable as

```js
['httpRequest', 'httpResponse', 'category'].includes(element.element)
```

or even

```js
VERY_SMART_CONSTANT_NAME.includes(element.element)
```

?

kylef: @honzajavorek We still support Node 4 (and browsers) with no transpilation so we can't use `includes` as it doesn't exist everywhere.

pksunkara (approved):

PR #199, opened 2018-11-13
Add basic fantasy land support to ArrayElement

kylef:
Implements small part of [Fantasy Land](https://github.com/fantasyland/fantasy-land) on ArrayElement.

The following conformance is supported:

- Functor (map)
- Semigroup (concat)
- Monoid (empty)
- Chain (chain)

Other concepts should be supported, but so far this I am only using the above in the OAS adapter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Minim Changelog
 
-## Master
+## 0.22.0
 
 ### Enhancements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Minim Changelog
 
+## Master
+
+### Enhancements
+
+- `ArrayElement` now conforms to parts of the [Fantasy
+  Land](https://github.com/fantasyland/fantasy-land) 3.5 specification.
+  `Functor`, `Semigroup`, `Monoid`, `Filterable`, `Chain`, and `Foldable` are
+  now supported.
+
 ## 0.21.1
 
 ### Bug Fixes
--- a/lib/primitives/array-element.js
+++ b/lib/primitives/array-element.js
@@ -312,7 +312,43 @@ var ArrayElement = Element.extend({
     return this.content.some(function (element) {
       return element.equals(value);
     });
-  }
+  },
+
+  // Fantasy Land
+
+  empty: function() {
+    return new this.constructor([]);
+  },
+
+  'fantasy-land/empty': function() {
+    return this.empty();
+  },
+
+  concat: function(other) {
+    return new this.constructor(this.content.concat(other.content));
+  },
+
+  'fantasy-land/concat': function(other) {
+    return this.concat(other);
+  },
+
+  'fantasy-land/map': function(transform) {
+    return new this.constructor(this.map(transform));
+  },
+
+  'fantasy-land/chain': function(transform) {
+    return this
+      .map(function (element) { return transform(element); }, this)
+      .reduce(function (a, b) { return a.concat(b); }, this.empty());
+  },
+
+  'fantasy-land/filter': function(callback) {
+    return new this.constructor(this.content.filter(callback));
+  },
+
+  'fantasy-land/reduce': function(transform, initialValue) {
+    return this.content.reduce(transform, initialValue);
+  },
 }, {}, {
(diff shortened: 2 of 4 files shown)


honzajavorek:
Delegating the review to @char0n as I think he's better suited to verify this kind of change.

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Minim Changelog
 
+## 0.22.0
+
+### Enhancements
+
+- `ArrayElement` now conforms to parts of the [Fantasy
+  Land](https://github.com/fantasyland/fantasy-land) specification. `Functor`,
```
char0n: Current version of fantasy-land is 3.5 (maybe worth mentioning)

lib/primitives/array-element.js
```diff
@@ -312,7 +312,48 @@ var ArrayElement = Element.extend({
...
+  },
+
+  'fantasy-land/map': function(transform, thisArg) {
+    return new this.constructor(this.map(transform, thisArg));
+  },
+
+  'fantasy-land/chain': function(transform) {
+    return this
+      .map(function (element) {
+        var result = transform(element);
```
char0n: According to spec of [chain](https://github.com/fantasyland/fantasy-land#chain) `transform must return a value of the same Chain`.  You're allowing `transform` to return multiple TypeReps which IMHO is not correct. `transform` should always return `ArrayElement` instances.

lib/primitives/array-element.js
```diff
@@ -312,7 +312,48 @@ var ArrayElement = Element.extend({
     return this.content.some(function (element) {
       return element.equals(value);
     });
-  }
+  },
+
+  // Fantasy Land
+
+  empty: function() {
```
char0n, line 319: you could use method shorhand for your methods:
```js
empty() {
   return new this.constructor([]);
},

File: test/primitives/array-element-fantasy-land.js [hunk 4]
```diff
@@ -0,0 +1,56 @@
+var expect = require('../spec-helper').expect;
+var minim = require('../../lib/minim').namespace();
+var R = require('ramda');
+
+var ArrayElement = minim.getElementClass('array');
+
+describe('ArrayElement', function() {
```
char0n: For tests I recommend using Fantasy-Land laws: https://github.com/fantasyland/fantasy-laws

Here is an example how to do it - https://github.com/char0n/ramda-adjunct/blob/master/test/fantasy-land/Identity.js. These laws grantee that your implementation implements various mathematical properties like e.g. reflexivity, transitivity, symmetry

lib/primitives/array-element.js
kylef: You're right, we allow using `[]` and `ArrayElement` interchangably in some of our Minim APIs making it simpler to use. I've made this stricter and only allowed the same type as you suggested to be more compilant with fantasy land.

lib/primitives/array-element.js
kylef: It's inconsistent with the rest of the code base (and some of the eslint style rules) so I won't do it here. Although we can move to that syntax (in separate PR with the rest of the code-base). Does it affect any browser/Node version support though?

char0n: Yep it's new feature from ES2015. Better to stay with the old syntax if your not using babel to transform to ES5.

test/primitives/array-element-fantasy-land.js
kylef: I like the idea, but I am struggling to get this done. There's a lot to grasp in fantasy-laws, along with its dependencies, jsverify and sanctuary.

I am wondering if perhaps we can proceed without fantasy-laws to get this out sooner and not block other work and perhaps we can sync up while I'm in Prague in few weeks @char0n. We could walk through implementing one of the tests together to help me understand how to fit it all together?

kylef: I've already improved the tests and moved away from Ramda to calling the fantasy land methods directly via `fantasy-land` package to catch typos in the method names.

char0n: Yep as discussed this is just fine. We'll sync on laws personally if that is something you want to pursue.

PR #209, opened 2019-01-15
refactor: Replace uptown with ES2015 classes

realityking:
Breaking change: Inheriting from a minim type now requires using native JS (class extends or prototypes) instead of uptown methods.

I'd like to propose removing the uptown dependency from minim and using ES2015 classes instead. This is a breaking change due to different semantics of ES2015 and uptown but it reduced code size & complexity and makes it easier for IDEs to understand the code (for autocomplete)

Perhaps the biggest downside is that users who want to use minim on the frontend will have to use Babel if they care about IE11 comparability. IMHO that's a fair trade off but others might feel otherwise.


smizell:
Please feel free to ignore me as I'm just adding a thought here :)

This is super cool. I think if you wanted to keep support for IE11 and non-Babel builds, `uptown` could be used to mutate all of the classes in a given namespace (I think). There is an `extendify` function that can take a class and add the `.extend` class method. I'm not sure if you all need this kind of support, but it could be a way to do this and keep older support. You could probably write a little function that takes a minim namespace and calls `.extendify` on each class.

https://github.com/smizell/uptown/blob/2fc1af0d3b7736fdc99fc980b34469f0ee598d31/lib/uptown.js#L38

Just a thought!

char0n:
@realityking If you run the code through the babel7, than you can get ES5 code that runs on every environment. And what get's distributed via npm will the babelified ES5 instead of original source code.

realityking:
@char0n Definitely an option. I saw that babel was removed from several Apiary packages so I wasn't sure if that was something that's wanted. Should I make that change? And what browsers should be supported?

honzajavorek:
I removed Babel from Dredd and Dredd Transactions, as they run solely on Node.js at this moment. There have been thoughts about making them browser-compatible, but that's long-term idea with a lot of prerequisites. Until that happens, Babel would be only an obstacle to the development.

realityking:
Alright. I'll make a PR to add babel to minim. Any comment on desired browser support?

pksunkara:
```
"browserslist": [
     ">0.25%",
     "not op_mini all",
     "ie 11"
 ]
```

realityking:
Added babel to the PR 🙂 

The browserified build more than 30KB smaller than before.

test/subclass-test.js
```diff
@@ -1,20 +1,22 @@
+'use strict';
```
pksunkara: Is this line needed?

.babelrc
```diff
@@ -0,0 +1,10 @@
+{
+  "presets": [
+    ["env", {
+      "targets": {
+        "browsers": [ ">0.25%", "not ie 11", "not op_mini all"],
```
pksunkara, line 5: Which way would be better? Like this or using `.browserlistrc`

pksunkara (commented):
Just a few comments.

test/subclass-test.js
realityking: Not anymore now that babel is around. Removed.

.babelrc
realityking: I prefer `.browserlistrc` but babel 6 doesn't support it. Babel 7 requires Node.js 6 and thus I ended up here 😆

pksunkara (approved):

PR #214, opened 2019-02-19
Enable airbnb style guides

kylef:
This change enables the airbnb style guides we're using elsewhere and fixes most of the problems.

As eslint requires Node >= 6, this introduces a breaking changed dropping Node 4 support. We've dropped Node 4 on most of our libraries so I don't see it as a problem.


src/primitives/element.js
```diff
@@ -138,12 +138,11 @@ class Element {
    * @param names {...elementNames}
    * @returns {ArraySlice}
    */
-  findRecursive() {
+  findRecursive(...elementNames) {
     if (arguments.length > 1 && !this.isFrozen) {
       throw new Error('Cannot find recursive with multiple element names without first freezing the element. Call `element.freeze()`');
     }
 
-    const elementNames = [].concat.apply([], arguments);
```
honzajavorek: Cool, this is much nicer now

test/object-slice-test.js
```diff
@@ -1,5 +1,6 @@
 const { expect } = require('./spec-helper');
 const minim = require('../src/minim');
+
 const { MemberElement } = minim;
 const { ObjectSlice } = minim;
```
honzajavorek: Wouldn't these make sense more as

```js
const { MemberElement, ObjectSlice } = minim;
```

src/namespace.js
```diff
@@ -178,7 +178,7 @@ class Namespace {
    */
   get elements() {
     if (this._elements === undefined) {
-      var name, pascal;
+      var name; var pascal;
```
honzajavorek: Shouldn't these be two lines?

src/object-slice.js
```diff
@@ -3,43 +3,33 @@ const ArraySlice = require('./array-slice');
...
 
   reject(callback, thisArg) {
     return this.filter(negate(callback, thisArg));
   }
 
   forEach(callback, thisArg) {
-    return this.elements.forEach(function (member, index) {
-      return callback(member.value, member.key, member, index);
-    }, thisArg);
+    return this.elements.forEach((member, index) => callback(member.value, member.key, member, index), thisArg);
```
honzajavorek: Wondering whether the fat arrow won't collide with the `thisArg`. Do we have this behavior tested so we're sure it works as intended even after this change?

(Relevant on multiple places in this commit)

src/primitives/element.js
```diff
@@ -186,7 +186,7 @@ class Element {
 
     if (!elementNames.isEmpty) {
       elements = elements.filter((element) => {
-        let parentElements = element.parents.map(element => element.element);
+        let parentElements = element.parents.map(e => e.element);
```
honzajavorek, line 189: I'm wondering whether `el` wouldn't be slightly better as `e` gets often used for errors or exceptions, but up to you if it's worth the change.

(Relevant to multiple places in the commit)

src/primitives/element.js
```diff
@@ -188,6 +188,7 @@ class Element {
       elements = elements.filter((element) => {
         let parentElements = element.parents.map(e => e.element);
 
+        // eslint-disable-next-line no-restricted-syntax
```
honzajavorek, line 191: why?

honzajavorek: oh I see, it returns from the iteration... okay then, this would require larger refactoring I guess

honzajavorek (approved):

src/namespace.js
kylef: Yes, but that comes in later rules.

In later commit the whole line goes away since we're using forEach so name vanishes and made pascal a const inside the forEach block.

src/object-slice.js
kylef: I naively though that would've been tested, but I can see that probably isn't the case. I'll test this behaviour and write missing tests.

kylef: Addressed in db73cc4.

honzajavorek (approved):

PR #216, opened 2019-02-22
Move to CircleCI

kylef:
We're using CircleCI in all other repos so I'd like to upgrade it here too for numerous reasons:

- More consistent and reliable build times
- Separate build statuses for each node / browser / linting
- Consistency with other repositories (api-elements.js / minim namespaces)

I also want to setup coverage reporting which I'll do in separate changeset.

Travis CI:

> Ran for 1 min 47 sec
> Total time 4 min 31 sec 

Circle CI:

> Total time 1 min 6 sec

NOTE, CircleCI time would actually be better but Node 6 is reliably slower than Node 8 and 10:

screenshot 2019-02-22 at 15 50 24

This pull request also removes browser testing from the `npm test` command, I think generally when developing (unless developing against the browser) you do not want `npm test` to run these additional tests which launch a browser and take more time. You can run `npm run test:browser` if you want them, but I think usually you do not. CI will always run them thought. I think in a follow up PR I'll see about testing on some WebKit-based browsers such as Chrome.


.eslintignore
```diff
@@ -1 +1 @@
-dist/*.js
+dist/
```
kylef, line 1: This was causing problems if you are linting after browser transpilation was done as there is other JS files inside dist/lib/.

package.json
```diff
@@ -10,12 +10,10 @@
     "build:browser": "browserify -d -s minim -t [ babelify ] -o dist/minim.js dist/lib/minim.js",
     "coverage": "istanbul cover _mocha -- -R spec --recursive --require babel-register",
     "coveralls": "coveralls <coverage/lcov.info",
-    "frontendtests": "karma start",
     "lint": "eslint .",
-    "nodejstests": "mocha -R spec --recursive",
     "precoveralls": "npm run coverage",
-    "prepublish": "npm run lint",
-    "test": "npm run nodejstests && npm run frontendtests"
+    "test": "mocha -R spec --recursive",
```
honzajavorek: I'm not fan of short options in written form as it's bad documentation, but it's not a sin I'd block the PR for

honzajavorek (approved):

package.json
kylef: Me neither, updated in 776a7db (where I just removed the -R option as spec is default) and move --recursive to mocha.opts

PR #217, opened 2019-02-22
Update dev dependencies and remove unused dependencies

kylef:
NOTE, Circle CI status will fail because of #216.

This leaves:

```shell
$ npm outdated
Package   Current   Wanted   Latest  Location
babelify    8.0.0    8.0.0   10.0.0  minim
lodash     4.17.4  4.17.11  4.17.11  minim
```

I don't want to touch lodash here as that is a normal dependency and it can have implications outside this project (dependency approvals etc) so I'll leave it as-is for now. Babelify I have left at v8 because there are some breaking changes in newer babel which require further changes here.


kylef:
Seems to break Node 6 / browser testing which I think will be worked around via #216, will investigate / wait for the linked PR first.

honzajavorek (approved):

PR #218, opened 2019-02-22
Update readme to remove uptown and fix incorrect code blocks

kylef:
To be honest, the state of this README isn't great and I'm unsure how many of the code examples are correct and there certainly is some methods that are implemented and not documented in the README. I'd like to remove the code-documentation from the README in the future and provide a separate websites (such as readthedocs) which documents the entire API generated from code so this stays in sync.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,8 +107,6 @@ var NewElement = StringElement.extend({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


kylef:
Again CI fails due to dependency on #216

honzajavorek (approved):

PR #230, opened 2019-06-20
Use ramda insead of lodash

kylef:
The biggest motivation was that we're using Ramda elsewhere which provides similiar functionality. This has a decent improvement in the overall bundle size, saving us 96K or 19% of the entire bundle size:

    $ du -h dist/*.js
    508K    dist/minim-with-lodash.js
    412K    dist/minim-with-ramda.js

### DIFF
--- a/lib/ArraySlice.js
+++ b/lib/ArraySlice.js
@@ -1,4 +1,4 @@
-const negate = require('lodash/negate');
+const complement = require('ramda/src/complement');
 
 // Coerces an a parameter into a callback for matching elements.
 // This accepts an element name, an element type and returns a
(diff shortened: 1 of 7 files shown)


kylef:
Added "do not merge", slighly concerned about performance difference for some of the change code paths, doing further analysis.

honzajavorek (dismissed):

kylef:
`element.id.equals` is now around a 4x slower operation on this branch.

kylef:
Another problem, `clone` from Ramda is not a deep clone (https://github.com/ramda/ramda/issues/2850). This behaviour would be breaking minim's interface and likely some of the OAS 2 and OAS 3 parsers.

PR #231, opened 2019-06-20
Simplify hasValue

kylef:
The `hasValue` function can be simplified using high order function `some`.


honzajavorek (approved):
TIL [some()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

PR #232, opened 2019-06-20
Serialise objects as O(N) instead of O(N^2)

kylef:
We're causing double loop of the elements in an object during serialisation which thus makes the serialisation slower. As we are looping over the keys, and then looking up the key from the content of an object again (thus causing secondary loop in `get`).


klokane (approved):
good catch

PR #240, opened 2019-09-10
Serialise undefined 0.6 array content as empty array

kylef:
The bug was discovered when doing conversions from a deserialised API Elements tree when using the JSON serialiser as the content for the element internally become undefined which means the serialiser wouldn't work.

Take the following example:

```js
const { ArrayElement } = require('minim');
const { JSON06Serialiser } = require('minim');

const array = new ArrayElement();
const serialiser = new JSON06Serialiser();
console.log(serialiser.serialise(array));

const request = new ArrayElement();
request.element = 'httpRequest';
request.content = undefined;
console.log(serialiser.serialise(request));
```

with incorrectly outputted:

```shell
$ node test.js
{ element: 'array' }
{ element: 'httpRequest' }
```

Whereas the example:

```js
const { ArrayElement } = require('minim');
const { JSON06Serialiser } = require('minim');

const array = new ArrayElement();
const serialiser = new JSON06Serialiser();
console.log(serialiser.serialise(array));

const request = new ArrayElement();
request.element = 'httpRequest';
request.content = undefined;
console.log(serialiser.serialise(request));
```

Correctly outputted:

```shell
$ node test.js
{ element: 'array' }
{ element: 'httpRequest', content: [] }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Minim Changelog
 
+## Master
+
+### Bug Fixes
+
+- Fixes a JSON 0.6 serialisation bug where httpRequest and similar array-based
+  elements with undefined content would be serialised with undefined content
+  instead of an empty array as content.
+
 ## 0.23.5 (2019-07-02)
 
 This release brings some performance improvements, namely to serialising with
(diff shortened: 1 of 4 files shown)


opichals:
👍

This fixes the issue.

pksunkara (approved):

